PR #31, opened 2015-01-29
move bash-completion to the correct place

### LINKED ISSUE #30: bash completion on Arch: possible packaging error
Hi,

I noticed the only file in `/usr/bash_completion.d/` is `downgrade`.

I found this strange, so when I tested bash completion with downgrade, it didn't worked.

However, after copying it to `/usr/share/bash-completion/completions/` (where almost all bash completions are on Arch), it worked. 

You might want to confirm this yourself first. I'll create a pull request and reference this issue from it.

thiagowfx:
Fixes #30.


pbrisbin:
Thank you. Will release a new version tomorrow.

PR #34, opened 2015-09-01
checking remote package signatures to fix MitM vulnerability

anthraxx:
it would be a great idea to also retrieve the remote signature
file of a package that is downloaded via pure unsecured http.
Without downloading the signature file it is very easy to use a
man-in-the-middle attack to deploy arbitrary files onto the victims
machine.

Maybe also consider creating a new release after this fix :smile:

### DIFF
--- a/downgrade
+++ b/downgrade
@@ -38,6 +38,10 @@ download_when_remote() {
       rm "$pkg"
       return 1
     fi
+    if ! curl --progress-bar "${url}.sig" > "${pkg}.sig"; then
+      rm "$pkg" "${pkg}.sig"
+      return 1
+    fi
   fi
 
   [[ -f "$pkg" ]] && printf "%s" "$pkg"


downgrade
```diff
@@ -38,6 +38,10 @@ download_when_remote() {
       rm "$pkg"
       return 1
     fi
+    if ! curl --progress-bar "${url}.sig" > "${pkg}.sig"; then
```
pbrisbin: Does just downloading the signature file cause it to be checked?

Are we assuming here that there will always be a `.sig` available?

anthraxx:
- it is always *.sig, yes... thats mandatory name.
- yes, depending on your configuration in /etc/pacman.conf
  the normal behavior is that signatures should be checked if available locally but its not enforced if installing from filesystem (thats why it worked at all, because it was manually downloaded over http)

pbrisbin:
Couldn't you use a MITM to modify the sig file too?

anthraxx:
No, because the key needs to be inside the default pacman keyring, thats not possible

pbrisbin:
Cool, thanks!

anthraxx:
@pbrisbin sure, you're welcome :smile:
Please also consider doing a new git-tag soon so this vulnerability is fixed for users....
Small heads-up: I'm considering to move your package into the main repos, would be nice to have a fixed version before doing so :smile:

anthraxx:
wow, you already did :+1: 
just ignore my previous post :yum:

PR #51, opened 2016-11-24
Implemented functionality to downgrade to a specific version

carmilso:
The option [-v <version>] allows to downgrade a package to a specific version. If the version is in cache too you must choose which of them you want (local or remote).

e.g.
    downgrade wpa_supplicant -v 2.3-1 (not installed, it downgrades it directly)
    downgrade wpa_supplicant -v 1:2.6-2 (currently installed, it shows two options)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,12 @@ Downgrade some packages, checking both local cache and the A.R.M.:
 $ downgrade foo bar
 ```
 
+Downgrade some package to a specific version:
+
+```
+$ downgrade <package> [-v version]
+```
+
 Downgrade a package, looking in only local cache:
 
 ```
@@ -24,7 +30,7 @@ Downgrade a package, looking in only the A.R.M.:
 $ NOCACHE=1 downgrade foo
 ```
 
-Downgrade a package, looking only in local cache, and favoring `su` over 
+Downgrade a package, looking only in local cache, and favoring `su` over
 `sudo` even when `sudo` is available:
 
 ```
--- a/downgrade
+++ b/downgrade
@@ -75,12 +75,14 @@ search_packages() {
   ((NOARM)) || \
     curl --fail --silent "$index" | sed '
       /.* href="\('"$1"'.*\(any\|'"$ARCH"'\)\.pkg\.tar\.xz\)".*/!d;
-      s||'"$index"'\1|g; s|+| |g; s|%|\\x|' | xargs -0 printf "%b"
+      s||'"$index"'\1|g; s|+| |g; s|%|\\x|' | xargs -0 printf "%b" | grep "\-$version"
+
+  local version_pattern="${version:-[0-9R]}"
 
   # shellcheck disable=SC2046
   ((NOCACHE)) || \
     find $( sed '/^#\?CacheDir *= *\(.*\)$/!d;s//\1/' "$PACMAN_CONF" ) \
-      -name "$1-[0-9R]*.pkg.tar.[gx]z"
+      -name "$1-$version_pattern*.pkg.tar.[gx]z"
 }
 
 sort_packages() {
@@ -95,6 +97,14 @@ output_package() {
          s|^/.*/\([^/]\+\)$|\1 ('"$(gettext 'local')"')|;' <<< "$2")"
 }
 
+check_packages() {
+  test "${#candidates[@]}" -gt 1 && present_packages "${candidates[@]}"
+}
+
+check_selection() {
+  test "${#candidates[@]}" -gt 1 && choice="$(read_selection "${candidates[@]}")"
+}
+
 main() {
   local term candidates choice pkg exit_code=0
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -24,7 +30,7 @@ Downgrade a package, looking in only the A.R.M.:
 $ NOCACHE=1 downgrade foo
 ```
 
-Downgrade a package, looking only in local cache, and favoring `su` over 
+Downgrade a package, looking only in local cache, and favoring `su` over
```
pbrisbin: Thanks for cleaning up the whitespace -- do you might moving those changes to their own commit?

downgrade
```diff
@@ -75,12 +75,14 @@ search_packages() {
   ((NOARM)) || \
     curl --fail --silent "$index" | sed '
       /.* href="\('"$1"'.*\(any\|'"$ARCH"'\)\.pkg\.tar\.xz\)".*/!d;
-      s||'"$index"'\1|g; s|+| |g; s|%|\\x|' | xargs -0 printf "%b"
+      s||'"$index"'\1|g; s|+| |g; s|%|\\x|' | xargs -0 printf "%b" | grep "\-$version"
```
pbrisbin: I'd use `grep -F` here to avoid things in `$version` being treated like a regex.

downgrade
```diff
@@ -103,8 +113,8 @@ main() {
   for term; do
     candidates=( $(search_packages "$term" | sort_packages) )
 
-    if present_packages "${candidates[@]}"; then
-      if choice="$(read_selection "${candidates[@]}")"; then
+    if check_packages || test "${#candidates[@]}" -gt 0; then
+      if check_selection || choice="${candidates[0]}"; then
```
pbrisbin: I'm having a little trouble following the intent of these nested, combined conditionals now. Can you explain what this code is doing? Then I might have some suggestions for clarifying the naming.

pbrisbin:
Hey, sorry I didn't see this until now! This is a great addition, and the patch looks good aside from a few comments. Let me know when you've addressed them.

downgrade
```diff
@@ -135,16 +145,18 @@ declare -a terms
 declare -a to_ignore
 declare -a to_install
 declare -a pacman_args
+declare -a version
 
 while [[ -n "$1" ]]; do
   case "$1" in
     --) shift; pacman_args=( "$@" ); break ;;
+    -v) shift; version=( "$1" ); break ;;
```
pbrisbin: You also shouldn't `break` out of the loop on `-v` or an invocation like `downgrade -v ... -- pacman-options` would never get to the pacman options.

pbrisbin:
Oh and one last thing: can you add a test (or a few) for this feature?

README.md
carmilso: Oh! It's done automatically by my vim configuration

downgrade
carmilso: You're right. Great improvement!

downgrade
carmilso: check_packages executes the function present_packages only if there is more than one possibility to download (e.g. if you search a incomplete version like 2.3 being versions 2.3-1 or 2.3-2 since the search of the version is a grep). Then, if there is more than one version (test "${#candidates[@]}" -gt 0) and after being potentially executed present_packages, it continues with the program.
check_selection executes the function read_selection only if there is more than one possibility to download (the same case than above). If there is only one possibility, then the choice is that version and it is not needed to ask the user the number of the version to download (check_selection || choice="${candidates[0]}")

downgrade
carmilso: Yes, in this version I assume that version couldn't be an array but if the pull request is accepted I can work on a version where each version corresponds to the corresponding package in the array. This works now basically to automate the process of downgrading a package when you know the specific version you want.
And you are right again about the break, bug...

carmilso:
How are you implementing the tests for this program? I have tried to downgrade some packages to a specific version and it has worked well (downgrading the package directly to that version)

pbrisbin:
This project uses [cram](https://bitheap.org/cram/) and you can find its current tests as examples in the [test](https://github.com/pbrisbin/downgrade/tree/master/test) directory, and see how they're run via the [CI config](https://github.com/pbrisbin/downgrade/blob/master/circle.yml).

The long and short of it is:

- Write a `foo.t` to hold your test(s)
- Each test is a command to run and expected output
- If the output doesn't match expected, it fails

downgrade
pbrisbin: I see. I think it's confusing to hide the "does thing if there's more than one option" behind these `check_` function. WDYT of writing this something like this:

```bash
if (( "${#candidates[@]" == 1 )); then
  choice=${candidates[0]}
else
  # do what it does now with present/read-selection/etc
  choice=...
end

# proceed, operating on `$choice`
```

It might require a little shuffling of the loop's `continue` and setting of `exit_code`, but I think it'd be worthwhile, and hopefully there are tests to ensure you don't break it.

pbrisbin:
This [blog post](https://pbrisbin.com/posts/mocking_bash/) of mine might be useful, as this is the project that motivated and still uses that strategy.

carmilso:
Excuse me, but I have not enough time now to realize all this modifications. I hope fixing it in few weeks

pbrisbin:
@carmilso I began a new branch based off yours, intending to just make those fixups, but I ended up taking things in a slightly different direction to accomplish this feature. Please let me know what you think: #52

carmilso:
I have revised it and I agree with all of that modifications from my proposal. I will try to collaborate with this improvement!

pbrisbin:
This has been implemented via #52 and released as 5.4.0. Thanks for the inspiration!

PR #66, opened 2018-01-20
 Use “+” indicator for currently installed and “-” indicator for packages installed in the past.

Thomaash:


downgrade
```diff
@@ -39,6 +39,12 @@ previously_installed() {
   ' "$PACMAN_LOG"
 }
 
+currently_installed() {
+  LC_ALL=C.UTF8 "$PACMAN" -Qi "$1" 2> /dev/null \
+  | tr '\n' '\t' \
+  | sed 's#^Name\s*: \([^\t]*\)\t.*Version\s*: \([^\t]*\).*$#\1-\2#'
```
pbrisbin: I think this way is a bit clear:

```sh
pacman -Qi downgrade | awk -F " : " '
  /^Name / { name=$2 };
  /^Version / { version=$2 };
  END { printf("%s-%s\n", name, version) }
'
```

```console
p -Qi downgrade | awk -F " : " '/^Name / { name=$2 }; /^Version / { version=$2 }; END { printf("%s-%s\n", name, version) }'
```

Unfortunately, it outputs ` - ` (not the empty string) for non-installed packages. I'm hopeful there's a way to address that if I keep thinking.

downgrade
```diff
@@ -127,14 +133,19 @@ sort_packages() {
...
-      *$pkg*)
-        indicator="*"
+  if [[ ! -z $current ]] && [[ $2 == *"$current"* ]]; then
+    # Currently installed
+    indicator="+"
+  else
+    for pkg in "${installed[@]}"; do
+      case "$2" in
+        *$pkg*)
+        indicator="-"
```
pbrisbin: Please maintain the existing style of indentation here:

```sh
case "$2" in
  *$pkg*)
    indicate="-"
    break
    ;;
```

downgrade
```diff
@@ -127,14 +133,19 @@ sort_packages() {
 output_package() {
   local pkg indicator=" "
 
-  for pkg in "${installed[@]}"; do
-    case "$2" in
-      *$pkg*)
-        indicator="*"
+  if [[ ! -z $current ]] && [[ $2 == *"$current"* ]]; then
```
pbrisbin: Please use `-n` for `! -z`.

I would quote `"$current"` and `"$2"` even though I don't think it's strictly needed (at least, ShellCheck didn't complain) because you're in `[[` and not `[`.

downgrade
```diff
@@ -176,6 +187,7 @@ main() {
 
   for term; do
     installed=( $(previously_installed "$term") )
+    current=( $(currently_installed "$term") )
```
pbrisbin: This makes current a single-element array (well it files it, you've `declare`d it as an array below). But then you treat it like a string with `-z` and `*$current*` above. I think you just want to use

```sh
```

Here, and drop the `declare` below.

test/bin/pacman
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+exec act-like /usr/bin/pacman "$@"
+
```
pbrisbin: Extra whitespace

test/present_packages/currently-installed.t
```diff
@@ -0,0 +1,23 @@
+  $ source "$TESTDIR/../helper.sh"
+
+Outputs an indicator next to previously installed packages
```
pbrisbin: :clap: THANK YOU for writing tests!

I think should be titled with "currently installed", right?

Thomaash:
I made quite a few copy paste errors, sorry about that. It should be fine now.

downgrade
```diff
@@ -39,6 +39,14 @@ previously_installed() {
   ' "$PACMAN_LOG"
 }
 
+currently_installed() {
+  LC_ALL=C.UTF8 "$PACMAN" -Qi "$1" 2> /dev/null | awk -F " : " '
+    /^Name / { name=$2 };
+    /^Version / { version=$2 };
+    END { if (name != "") printf("%s-%s\n", name, version) }
```
pbrisbin: Nice

downgrade
```diff
@@ -127,14 +135,19 @@ sort_packages() {
...
+    # Currently installed
+    indicator="+"
+  else
+    for pkg in "${installed[@]}"; do
+      case "$2" in
+        *$pkg*)
+          indicator="-"
+          break
+          ;;
     esac
```
pbrisbin: I think this needs to be shifted two spaces, to align with `case` above.

Thomaash: Yeah I somehow overlooked this line, sorry about it.

PR #67, opened 2018-01-26
Czech translation

Thomaash:


pbrisbin:
Thanks! I think the history here is off for some reason, can you look into that -- maybe an old branch?

PR #77, opened 2019-03-18
Organize package info into a table

### LINKED ISSUE #49: tabular output
Currrently, downgrade outputs the available versions as (e.g.)
```
   1) mercurial-4.0-1-x86_64.pkg.tar.xz (remote)
   2) mercurial-4.0-1-x86_64.pkg.tar.xz (local)
   3) mercurial-3.9.2-1-x86_64.pkg.tar.xz (remote)
   4) mercurial-3.9.2-1-x86_64.pkg.tar.xz (local)
   5) mercurial-3.9.1-1-x86_64.pkg.tar.xz (remote)
   6) mercurial-3.9-1-x86_64.pkg.tar.xz (remote)
   7) mercurial-3.8.3-1-x86_64.pkg.tar.xz (remote)
   8) mercurial-3.8.2-1-x86_64.pkg.tar.xz (remote)
   9) mercurial-3.8.1-2-x86_64.pkg.tar.xz (remote)
```
Note how the differing version lengths make the whole thing slightly tricky to read.

A minor improvement would be to output
`$index) $pkgname <TAB> $pkgver <TAB> $pkgrel <TAB> $arch <TAB> ($local_or_remote)`

Thomaash:
Fixes #49.
[Sample output.txt](https://github.com/pbrisbin/downgrade/files/2980863/Sample.output.txt)

### DIFF
--- a/downgrade
+++ b/downgrade
@@ -25,8 +25,11 @@ currently_installed() {
   '
 }
 
+# <package name> [package path] …
 present_packages() {
   local i=1
+  local pkgname="$1"
+  shift
 
   (($#)) || return 1
 
@@ -34,8 +37,8 @@ present_packages() {
   printf '\n\n'
 
   for entry; do
-    output_package "$((i++))" "$entry"
-  done
+    output_package "$((i++))" "$entry" "$pkgname"
+  done | column -s '	' -t -R 2,4,6
 
   printf '\n'
   gettext 'select a package by number: '
@@ -111,15 +114,16 @@ sort_packages() {
     pacsort -f -t '|' -k 1 | cut -d '|' -f 2-
 }
 
+# <number> <path> <package name>
 output_package() {
-  local pkg indicator=" "
+  local number="$1" path="$2" pkgname="$3" pkg indicator=" " location version epoch arch release
 
-  if [[ -n "$current" ]] && [[ "$2" == *"$current"* ]]; then
+  if [[ -n "$current" ]] && [[ "$path" == *"$current"* ]]; then
     # Currently installed
     indicator="+"
   else
     for pkg in "${installed[@]}"; do
-      case "$2" in
+      case "$path" in
         *$pkg*)
           indicator="-"
           break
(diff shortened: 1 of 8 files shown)


downgrade
```diff
@@ -128,10 +132,48 @@ output_package() {
...
-         s|^/.*/\([^/]\+\)$|\1 ('"$(gettext 'local')"')|;' <<<"$2"
+  # Remote or local file
+  if [[ $path =~ ^/ ]]; then
+    location="$(gettext 'local')"
+  else
+    location="$(gettext 'remote')"
+  fi
+
+  # Remove the path, package name and extensions
+  # [epoch:]version-release-arch
```
pbrisbin: This bit is the only thing that concerns me. I think it could be done more simply, and I'm not sure it handles all possible kinds of `version` strings correctly. Before addressing that, we would need to get it wrapped in a unit test.

Can you please extract this logic to a dedicated function that accepts the path values on `stdin` and outputs separate words for the parts it's "parsing" out:

```sh
# Pick a better name please ;)
break_up_path()
  # your implementation here
}

# how you would use it:
read -r epoch version release arch < <(break_up_path <<< "$path")
```

Then wrap it in a unit test for all kinds of nasty `path` values?

downgrade
```diff
@@ -25,17 +25,20 @@ currently_installed() {
...
   (($#)) || return 1
 
   gettext 'Available packages:'
   printf '\n\n'
 
   for entry; do
-    output_package "$((i++))" "$entry"
-  done
+    output_package "$((i++))" "$entry" "$pkgname"
+  done | column -s '	' -t -R 2,4,6
```
pbrisbin: What does this do?

Thomaash: This creates space indented table (output_package just divides cells with tabs).
I added a comment to it.

downgrade
Thomaash: I skimmed through some repos and (apart from a small bug that is fixed now) it seems to work just fine.

pbrisbin: Gotcha. I'm pretty sure that if a package contained a `-` in its version string, this would break. I'm not sure if that's legal or not, but I can't find an existing example, so I'm good to not worry about it for now.

Thomaash:
CircleCI tests fail because there's no column command. On Arch this command is provided by util-linux package which is in base and base-devel goups and is also required by mkinitcpio and systemd among others, so almost everybody already has it anyway. But I don't know what can be done about it in CircleCI.

pbrisbin:
You just have to find out to install it in debian and install it like we do `gettext`: https://github.com/pbrisbin/downgrade/blob/master/.circleci/config.yml#L18

Thomaash:
I give up, this is not worth the effort. Debian removed column from util-linux and seems to offer only different one based on column from BSD which works differently and doesn't have a lot of functionality.

pbrisbin:
Wow that's annoying. The other option is to add `test/bin/column` which will be found first. You can make that be a simple `cat` our something else to indicate it is used. This is how we avoid needing `pacman` on CI.

Thomaash:
Hmm, I didn't think of that.

Thomaash:
Finally, everything's green ☺.

Thomaash:
You're welcome.

Issue #83, opened 2019-12-04
Downgrading hard + version-bound dependencies


atreyasha:
Hello @pbrisbin,

Thank you for making this awesome tool, it's really useful! Going back to issue #19, I think it might be really useful to have dependency management while downgrading a package. I understand you mentioned that you don't want to add it, but would it be possible to see how this works if I submit a PR?

If so, it would be great if you could advise me which area of the code you can envision this coming in; that would make it slightly quicker :)

pbrisbin:
Hi there, as I mentioned in the original issue:

> I'll gladly review any PRs that make this situation easier though

So do feel free to explore this.

> it would be great if you could advise me which area of the code you can envision this

Hmm, I'm really not sure. It'll be a large change that will take some time to get done (I think) so an approach that keeps things mostly isolated from the existing code would provide the best chance of it getting reviewed and merged.

I _imagine_ the process would have to work something like:

1. User is prompted with versions available for the original argument package
2. User makes her selection
3. _Somehow, we find that version's `PKGBUILD` or `SRCINFO` or whatever_
4. _Insert dependency resolution function that uses this to (recursively) spit out more `package-version`s that need processing_
5. The user's first argument, and all these new ones, are passed into the actual install-ignore function

(3) and (4) seems like the biggest risks, so I might begin by simply working on those in total isolation. If that can't be solved, there's no use in going any further.

**As an alternative**:

Downgrade's usage is flexible enough to be used as part of this process:

- It accepts full `{package}-{version}` arguments and will downgrade directly to that version without any prompting
- It accepts _multiple_ arguments

Therefore, if some external tool could sort out the dependency graph of a multi-package downgrade, you could simply pass that tool's output to `| xargs downgrade`.

That approach is super attractive to me because it's very UNIX and leverages downgrade without bloating it. But the major downside is that downgrade is also good at taking a package name and prompting you based on Cache and ARM as to what you might want to install. That makes this tricky.

What about something like this:

```console
% downgrade firefox --print | fzf | downgrade-dependencies | xargs downgrade
```

1. `downgrade --print` dumps the available `{package}-{version}`s in Cache or ARM
2. `fzf` now handles making your selection and printing that one out
3. `downgrade-dependencies` is a separate executable that does the heavy lifting and prints out more packages
4. Then `xargs downgrade` comes back to install and handle `IgnorePkgs`

[...]

PR #87, opened 2020-01-04
Downgrading dependencies of user-defined packages

atreyasha:
As per issue #83, here is a proposed PR to downgrade dependencies. Essentially, after downgrading the user-defined packages, it moves to identify reverse dependencies using `pactree` and then re-uses the previously defined functions in `downgrade` to correspondingly downgrade dependencies. This is summarized in the new function `downgrade_dependencies`.

All tests were passed using `make test`.

Some caveats:

1. We would need to add `pacman-contrib` (which offers `pactree`) as a dependency to `downgrade`

2. As per my knowledge, the original `sudo pacman -U` command should already take care of dependencies if they are explicity stated with version numbers. Otherwise this PR provides additional tools to manually downgrade dependencies.

3. It also involves a running loop for each main package, which allows the user to downgrade dependencies one-by-one.

    Furthermore, I added a `q` option to quit this loop in case the user does not want to downgrade dependencies anymore. This required an additional `read_selection_dependencies` function so as to not conflict with the original `read_selection` function.

See [here](https://github.com/atreyasha/downgrade/blob/master/doc/quick.gif) for sample gif. Give it a try with some packages and let me know your feedback.


downgrade
```diff
@@ -274,6 +344,7 @@ cli() {
   main "${terms[@]}"
   sudo pacman -U "${pacman_args[@]}" "${to_install[@]}"
   prompt_to_ignore "${to_ignore[@]}"
+  downgrade_dependencies "${terms[@]}"
```
pbrisbin: I think downgrading dependencies separately from the packages that depend on them could fail due to version bounds issues (no matter what order you do them in). I think we need all installations for a given initial argument to be passed at-once to `pacman -U`.

I also think that'd be a better design (the `unset`s are a smell, IMO). Something like this:

```sh
add_dependencies() {
  # do all the logic you're currently doing, but don't actually install, just
  to_install=( "$newly_selected_dependency" "${to_install[@]}" )
  to_ignore=( "$newly_selected_dependency" "${to_ignore[@]}" )
}

main "${terms[@]}"
add_dependencies "${terms[@]}"
sudo pacman -U "${pacman_args[@]}" "${to_install[@]}"
```

(Really, `add_dependencies` should probably be a part of `main`, but I just showed it this way because it's a bit easier to explain what I'm thinking.)

WDYT?

atreyasha: Thanks for the feedback. I agree that this is just a short-term solution, and that there are perhaps better ways of doing this whole process. For me, the main idea would be that the user can directly see all dependencies and decide possibly whether to downgrade some dependencies.

About your comment above, here are my thoughts:

1. If version bounds are provided as a necessity in a package, then `sudo pacman -U` would also downgrade the dependencies as per the documentation in `man pacman`.

```
-U, --upgrade
	 Upgrade or add package(s) to the system and install the required
	 dependencies from sync repositories. Either a URL or file path can be
	 specified. This is a “remove-then-add” process. See Upgrade Options
	 below; also see Handling Config Files for an explanation on how pacman
	 takes care of configuration files.
```

2. So I think the problems arise in cases where version bounds are not listed. I encountered such issues with packages such as `tensorflow` and `python-tensorflow`, in which both packages are linked together and the former is a dependency of the latter. However, their version bounds are not provided in the `pacman -Qi` documentation, which is why the user would have to manually downgrade them.

3. In regards to what you suggested, I think this approach is good. However, I can imagine a (probably rare) scenario where the selected downgraded dependencies would have some conflict with the downgraded main package. This would lead to some error in the `sudo pacman -U` process and would result in the script failing. The user would then have to re-run `downgrade` and figure things out by trial and error. Alternatively, we would have to catch the error and return to the downgrade process, which would result in more modifications, which is probably not desired.

The advantage of the existing approach is that the downgrading of non-version-bound dependencies happens after the downgrading of the main package. This would mean the user would remain in the running "downgrade-dependency" loop and could downgrade dependencies by trail-and-error without leaving the loop, since any error message would not break the loop.

But these are just my two cents. Tbh I think both approaches are fine. I would defer to your opinion on this, since you have probably encountered more user-feedback about the packages and issues related to them. Just let me know, and I can make those changes if need be.

pbrisbin: > For me, the main idea would be that the user can directly see all dependencies and decide possibly whether to downgrade some dependencies.

I think this is true with either approach, right?

> version bounds are provided as a necessity in a package, then sudo pacman -U would also downgrade the dependencies

I'm thinking that would only work for in-cache dependencies, not those in the remote archive.

> So I think the problems arise in cases where version bounds are not listed

I didn't realize that was common.

> The advantage of the existing approach is ... the user would remain in the running "downgrade-dependency" loop and could downgrade dependencies by trail-and-error without leaving the loop

This is convincing; I'm on board now :+1:

> I would defer to your opinion on this, since you have probably encountered more user-feedback about the packages and issues related to them

I wouldn't do that :wink: , this comment tells me you have thought about this particular problem more than I have and I actually would like to defer to you.

downgrade
```diff
@@ -54,6 +54,27 @@ read_selection() {
...
+  fi
+}
+
+enumerate_array() {
+  local i=1
+  local array=("$@")
+  for entry in "${array[@]}"; do
+    printf "%s)\t%s\n" "$((i++))" "$entry"
+  done
+  printf "q)\texit\n"
```
pbrisbin: I think we need a translation for `exit`.

downgrade
```diff
@@ -81,6 +102,56 @@ prompt_to_ignore() {
...
+    if [ ${#dep_array[@]} -eq 0 ]; then
+      return 0
+      # if dependencies are detected, enter loop
+    else
+      # while loop to downgrade dependencies
+      eval_gettext "review and downgrade dependencies of \$pkg? [y/n] "
+      read -r ans
+      if [[ "${ans,,}" == $(gettext 'y')* ]]; then
+        while true; do
+          printf 'Dependency graph:\n\n'
```
pbrisbin: Missing `gettext`.

downgrade
atreyasha: > > For me, the main idea would be that the user can directly see all dependencies and decide possibly whether to downgrade some dependencies.
> 
> I think this is true with either approach, right?

Ah yes, this is true. I was thinking out loud (habit)

> > version bounds are provided as a necessity in a package, then sudo pacman -U would also downgrade the dependencies
> 
> I'm thinking that would only work for in-cache dependencies, not those in the remote archive.
>
> > So I think the problems arise in cases where version bounds are not listed
> 
> I didn't realize that was common.

As per my experience with downgrading packages, I rarely encounter dependencies which have been documented with version bounds. Most of the time they are documented only with the name of the dependency package. And in many ways, I can imagine why. It incorporates so much more work/scrutiny for a developer to investigate and explicitly state version bounds for all dependencies, so perhaps they just do it for a few critical ones where critical functionality is dependent on a certain version bound.

In regards to this, I did a bit of investigating. One set of packages I found with version bound dependencies is `vim` and `vim-runtime` (verify with `pacman -Qi vim`). I tried running the following command:

```shell
$ downgrade vim
```

After selecting a `vim` package some steps back in time, I come across an unresolvable dependency issue related to `vim-runtime`. Maybe you can try it out too. But essentially this would imply that `sudo pacman -U` is not taking care of the `vim-runtime` dependency, as we would expect it to (regardless of local or remote).

I tried adding the following `pacman` argument to ignore dependencies:

```shell
$ downgrade vim -- --nodeps
```

The installation of an older version of `vim` pushed through, but then naturally vim-specific errors arose because of the wrong `vim-runtime` version. So this is not going to work.

One alternative I can foresee is that if there are unresolvable dependencies, we can catch this error and advise the user to uninstall the package using `pacman -Rns` and then reinstalling an older version using `downgrade`.

Alternatively, we could provide a prompt within `downgrade` to do this for the user, but this might be dangerous for critical packages such as kernel related ones where a user might delete kernel packages by mistake by using this hypothetical feature.

What do you think about this issue related to version-bounded dependencies?

> > The advantage of the existing approach is ... the user would remain in the running "downgrade-dependency" loop and could downgrade dependencies by trail-and-error without leaving the loop
> 
> This is convincing; I'm on board now 
> 
> > I would defer to your opinion on this, since you have probably encountered more user-feedback about the packages and issues related to them
> 
> I wouldn't do that  , this comment tells me you have thought about this particular problem more than I have and I actually would like to defer to you.

Nice :wink:, always good to collaborate.

> I'm on board with shipping this, if you're happy with it. If it's OK with you, I'd like to ship it disabled by default first though. Can you add a new environment variable, `DOWNGRADE_DEPENDENCIES`, default to `0` and only perform the check if `1`? Please also document this in the man-page markdown (I can handle rendering it on release).
> 
> Thanks!

I agree in making it optional at first, so we can test it out with users and catch/fix any issues. I will make those changes requested below and will update the markdown document.

On another note, I have two points for feedback.

1. I find the approach of parsing `pacman` arguments a bit different from my experience with other command-line tools, where the arguments come after a `--` delimiter and also after specifying the package name(s). What do you think about having this the other way around, such as:

**From**:
```shell
$ downgrade vim -- --nodeps
```

**To**:
```shell
$ downgrade --nodeps vim
```

2. Additionally, I find that we could convert the declaration of local environmental variables (such as `DOWNGRADE_FROM_ALA=0`) into non-pacman command line arguments. This would also fit in with the look-and-feel of other argument parsing command line tools.

    As an example, we could use the option `--ala` to declare only updating from the ALA. This option could come after the declaration of the package name(s) such that it does not conflict with `pacman` arguments. So for example:

**From**:
```shell
$ DOWNGRADE_FROM_CACHE=0 downgrade vim -- --nodeps
```

**To**:
```shell
$ downgrade --nodeps vim --ala
```

This might help new users, who could then use `downgrade` with their intuitive knowledge of other command-line tools, without having to understand the different interface of `downgrade`. I can gladly incorporate these features.

Sorry if this is going all over the place. Just thought it would be good to incorporate some changes which might help.

atreyasha: Okay, so I figured out the issue related to `vim`, which I mentioned above. In order to successful downgrade `vim` with its version-bound dependency `vim-runtime`, one would need to do `sudo pacman -U` with both `vim` and `vim-runtime` packages simultaenously to resolve the circular dependency (source: https://bbs.archlinux.org/viewtopic.php?id=122791).

If we integrate this as part of `downgrade`, we would need to change a bit of the `main` function. So we we would need to look through the cached or downloaded `.PKGINFO` file for the desired package, find any explicit version-bound dependency and install it alongside the original package in `sudo pacman -U`. This corresponds to exactly what you suggested before, and this could correspond to a fix for downgrading packages with explicit version-bound dependencies. I can also gladly incorporate this into the `main` function.

What do you think?

pbrisbin: I'm not going to reply point-by-point because it was a lot, *but* don't think I don't appreciate your thoughts, I read and agree with basically all of it :smile:

So here's what' I think is worth some continued discussion:

If doing what I suggested originally, so packages and their dependencies are passed to `pacman -U` together works for versioned and un-versioned dependencies, in cache or ALA, then I think that's the way to go. Don't bother looking at `.PKGINFO` to decide how to behave, just do it that way all the time.

If it *doesn't* work for both scenarios, I'd want to think a little bit more about it. It sounds like a lot of complexity, which makes me more reluctant to add it in `downgrade` proper, vs going all-in on a more pipe-line UX that allows this all to live in an isolated tool.

I'm fine with adding options to the script. Would you want to remove the env var support, or have them still alter behavior when no options are given?

My other too, `aurget` will capture any unknown arguments and consider them as `makepkg` arguments. What do you think about doing that here for the pacman arguments? Then it doesn't matter if they're passed before or after the package name or whatever. If it's unknown to downgrade, it's passed to `pacman`.

atreyasha: > I'm not going to reply point-by-point because it was a lot, _but_ don't think I don't appreciate your thoughts, I read and agree with basically all of it 

Thank you man :smile:, sorry I got into a funk after finally getting some free time to look through everything.

> So here's what' I think is worth some continued discussion:
> 
> If doing what I suggested originally, so packages and their dependencies are passed to `pacman -U` together, works for versioned and un-versioned dependencies, in cache or ALA, then I think that's the way to go. Don't bother looking at `.PKGINFO` to decide how to behave, just do it that way all the time.
> 
> If it _doesn't_ work for both scenarios, I'd want to think a little bit more about it. It sounds like a lot of complexity, which makes me more reluctant to add it in `downgrade` proper, vs going all-in on a more pipe-line UX that allows this all to live in an isolated tool.

I agree with these two points. More on this below.

> I'm fine with adding options to the script. Would you want to remove the env var support, or have them still alter behavior when no options are given?

Ideally speaking, I would like to remove explicit environmental variable declaration from `downgrade`. So I would like to have everything done through command line arguments.

That being said, `downgrade` was made quite some time back and perhaps some users already have the environmental variables set up as part of scripts. So maybe we can include an exception (at least for this pull request and the corresponding early release) to handle environmental variables. But I would suggest that we already change the documentation and the `man` entry to remove environmental variables and replace them with command line arguments.

> Another script of mine, `aurget`, will capture any unknown arguments and consider them as `makepkg` arguments. What do you think about doing that here for the `pacman` arguments? Then it doesn't matter if they're passed before or after the package name(s) or whatever. If it's unknown to `downgrade`, it's passed to `pacman`. I'm also honestly find to remove this feature, I'm not sure how useful it is.

That is great to hear! Coincidentally, I was googling for such `pacman` tools. Funny how it was just here and I missed it :laughing:

**Summary**

Okay so now to walk the walk, here are the summarized changes I will focus on:

1. Add versioned/unversioned dependencies handling with `sudo pacman -U` instead of having this at the end, ie. have the downgrading process done together in one step as per your initial suggestion.

2. Fix some of terms and `gettext` issues that you tagged below.

3. Replacing environmental variables with command-line arguments using `aurget` to help filter these arguments.

4. Adding a helper function for the command-line arguments and updating the documentation and `man` entry for these.

I will try to be generous with my changes, so we can just pick-and-choose the ones that we like. I am quite busy now with work, so give me a few weeks and I will revert with something concrete.

If you have any other suggestions, just let me know.

pbrisbin: > we can include an exception (at least for this pull request and the corresponding early release) to handle environmental variables. But I would suggest that we already change the documentation and the man entry to remove environmental variables and replace them with command line arguments.

:+1:

> Summary

1. :+1:
2. :+1:
3. :+1: with the minor clarification that you'll be making `dowgrade` _behave like_ `aurget`; there's no way to *use* `aurget` for this. You probably meant that, but I just wanted to clarify
4. :+1:

> I will try to be generous with my changes, so we can just pick-and-choose the ones that we like. 

Awesome! Separate commits (or even better, PRs) sounds amazing. If something's a clean diff and low-impact, I'll merge right away. Then we can be more focused with the more complicated parts.

> I am quite busy now with work, so give me a few weeks

No rush!

atreyasha: Hi @pbrisbin,

Hope you are doing well. Unfortunately, I have been super swamped with work/studies, but all of this mess should be coming to an end in a month or two. I will finish up this PR once that is over; sorry again for the wait.

pbrisbin: No pressure from my end, good luck with work/studies!

PR #89, opened 2020-01-12
Fix downgrade not finding packages with .zst extension

Ferdi265:
This PR fixes `downgrade` for packages with the extension `.zst`, which is the new default package compression extension.

This PR also changes the package extension matching regex for ALA and CACHE resemble each other more.

### DIFF
--- a/downgrade
+++ b/downgrade
@@ -96,7 +96,7 @@ search_packages() {
 
   if ((DOWNGRADE_FROM_ALA)); then
     curl --fail --silent "$index" | sed '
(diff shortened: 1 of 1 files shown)


pbrisbin:
Thanks! Will merge and release on Monday.

pbrisbin:
The new default package compression extension.

Also changes the package extension matching regex for ALA and CACHE
resemble each other more, which is required since the new extension is more
than a one-letter difference (so the find glob can't work).

PR #93, opened 2020-03-09
Use less restrictive package file patterns by fully matching the filename

### LINKED ISSUE #92: Not showing packages with versions starting with lowercased r
It fails to show package files from my cache for a particular package and I find this:

https://github.com/pbrisbin/downgrade/blob/1f4a7459a0f3593c9b6c4e8c11d41751e2c3f959/downgrade#L91

I have no packages starting with an uppercased `R` but several of my -git packages start with a lowercased `r` (there are a few `r*` in [community] too). Also lshw starts with `B`, 0ad starts with `a`, bettercap-caplets starts with `v`, coxeter starts with `git`. This pattern seems too restrictive.

lilydjwg:
This fixes #92.

The old behavior is preserved if a version-like string is passed in because there isn't a way to decide if there is a (partial) version or not.

There are still scenarios that downgrade doesn't work but people should find that downgrade works for strictly more package names without any new issues.

### DIFF
--- a/downgrade
+++ b/downgrade
@@ -82,22 +82,23 @@ prompt_to_ignore() {
 }
 
 search_packages() {
-  local name version index
+  local name version pkgfile_re index
 
   if [[ "$1" =~ -([0-9R].*)$ ]]; then
-    version=${BASH_REMATCH[1]}
+    version="${BASH_REMATCH[1]}"
     name=${1%-$version}
+    version="$version.*"
   else
-    version='[0-9R]'
+    version="[^-]+-[0-9.]+"
     name=$1
   fi
-
+  pkgfile_re="$name-$version-(any|$DOWNGRADE_ARCH)\\.pkg\\.tar\\.(gz|xz|zst)"
   index="$DOWNGRADE_ALA_URL/packages/${name:0:1}/$name/"
 
   if ((DOWNGRADE_FROM_ALA)); then
-    curl --fail --silent "$index" | sed '
-      /.* href="\('"$name-$version"'.*\(any\|'"$DOWNGRADE_ARCH"'\)\.pkg\.tar\.\(gz\|xz\|zst\)\)".*/!d;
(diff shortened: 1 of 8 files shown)


lilydjwg:
A test fails because it provides fake package names that pacman doesn't actually use....

pbrisbin:
Are they not representative of actual packages? Happy to make them more realistic if that's the problem, but I'm not sure why `foo-1.1` wouldn't be a realistic name.

lilydjwg:
Because all generated package names have architecture part in their names. I'm going to sed them, hoping it works.

lilydjwg:
It didn't work :cry: . I'll try to fix and test them locally tomorrow.

pbrisbin:
Ah, yes I see. As some extra context, here's my thoughts on the failures:

1. extract, It works on a package with only version

   If you're right, and we'd never encounter a package without arch, feel free to remove this test.

2. extract, It works on a package with epoch:version-release & It works on a package with only version-release

   Having an arch or not doesn't matter to this tests. Feel free to update the inputs or the assertions.

3. search-packages with foo

   I can't immediately see the problem here. This may be a real bug, as I'd expect it to work.

4. search-packages with foo-completions

   This is exactly the bug I was mentioning originally, where searching for `foo` finds packages `foo-x`. This is probably the only failure we should discuss more, to decide if it's a false-positive (unrealistic scenario) or just something we need to accept in order to not have the bug you reported.

lilydjwg:
Sorry for the delay. I didn't use shell test frameworks before and thought it would take some signification time, so I hesitated but it turned out that cram is pretty to get start.

The failures are all caused by the fact that I added arch part. The first two are tests that needs update accordingly, and the latter are caused by the stricter filename matching that will parse filenames without release wrongly. I count `-`s to determine which part is what so a missing part causes problems.

pbrisbin:
Thanks! Will release (hopefully) later today.

PR #96, opened 2020-04-13
Adding command-line arguments for downgrade

atreyasha:
Hi @pbrisbin, finally got some time to work on the pull requests. In this PR, I added CLI arguments that can be parsed and used either directly in `downgrade` or by `pacman`. I used the same flow as per `aurget`, as per your recommendation. The previous environmental variables will still work if they are fed in.

Additionally, I added a helper/usage script that can be seen by executing `downgrade --help`. This script also pops up when there are error messages encountered. I added error-based message popups for cases where arguments are wrongly supplied, when packages are not supplied or when supplied packages are missing in the supposed caches/servers

If these changes are fine, I can update the readme and the man-page documentation.

### DIFF
--- a/doc/.gitignore
+++ b/doc/.gitignore
@@ -1,0 +1,2 @@
+todos.org
+Peek*
--- a/doc/.gitignore
+++ b/doc/.gitignore
@@ -1,2 +1,0 @@
-todos.org
-Peek*
--- a/downgrade
+++ b/downgrade
@@ -213,7 +213,7 @@ process_term() {
   else
     if present_packages "$term" "${candidates[@]}"; then
       choice=$(read_selection "${candidates[@]}")
-      if [ -z "$choice" ]; then
+      if [[ -z "$choice" ]]; then
         {
           gettext "Invalid choice"
           echo
--- a/downgrade
+++ b/downgrade
@@ -214,7 +214,10 @@ process_term() {
     if present_packages "$term" "${candidates[@]}"; then
       choice=$(read_selection "${candidates[@]}")
       if [ -z "$choice" ]; then
-        { gettext "Invalid choice"; echo; } >&2
+        {
+          gettext "Invalid choice"
+          echo
+        } >&2
       fi
     fi
   fi
@@ -280,9 +283,9 @@ cli() {
           PACMAN="$1"
         else
           {
-          gettext "Missing --pacman argument"
-          echo
-          gettext "$(usage)"
+            gettext "Missing --pacman argument"
+            echo
+            gettext "$(usage)"
           } >&2
           exit 1
         fi
@@ -293,9 +296,9 @@ cli() {
           PACMAN_CONF="$1"
         else
           {
-          gettext "Missing --pacman-conf argument"
-          echo
-          gettext "$(usage)"
+            gettext "Missing --pacman-conf argument"
+            echo
+            gettext "$(usage)"
           } >&2
           exit 1
         fi
@@ -306,9 +309,9 @@ cli() {
           DOWNGRADE_ARCH="$1"
         else
           {
-          gettext "Missing --arch argument"
-          echo
-          gettext "$(usage)"
+            gettext "Missing --arch argument"
+            echo
+            gettext "$(usage)"
           } >&2
           exit 1
         fi
@@ -319,9 +322,9 @@ cli() {
           DOWNGRADE_ALA_URL="$1"
         else
           {
-          gettext "Missing --ala-url argument"
-          echo
-          gettext "$(usage)"
+            gettext "Missing --ala-url argument"
+            echo
+            gettext "$(usage)"
           } >&2
           exit 1
         fi
@@ -355,9 +358,9 @@ cli() {
 
   if ((!"${#terms[@]}")); then
     {
-    gettext "No packages provided for downgrading"
-    echo
-    gettext "$(usage)"
+      gettext "No packages provided for downgrading"
+      echo
+      gettext "$(usage)"
     } >&2
     exit 1
   fi
--- a/downgrade
+++ b/downgrade
@@ -211,7 +211,7 @@ process_term() {
     choice=${candidates[0]}
   else
     if present_packages "$term" "${candidates[@]}"; then
-      choice=$(read_selection "${candidates[@]}")
+      choice=$(read_selection "${candidates[@]}") || gettext "Invalid choice"
     fi
   fi
 
@@ -233,10 +233,8 @@ main() {
     # shellcheck disable=SC2207
     installed=($(previously_installed "$term"))
     current=$(currently_installed "$term")
-    process_term "$term" || {
-      exit_code=1
-      gettext "No package with name $term found"
-    }
+    [[ -z "$current" ]] && gettext "No package with name $term found"
+    process_term "$term" || exit_code=1
   done
 
   return $exit_code
--- a/downgrade
+++ b/downgrade
@@ -268,24 +268,48 @@ cli() {
   while [[ -n "$1" ]]; do
     case "$1" in
       --pacman)
-        [[ -n "$2" ]] || (gettext "Missing --pacman argument" && echo && gettext "$(usage)" && exit 1)
-        shift
-        PACMAN="$1"
+        if [[ -n "$2" ]]; then
+          shift
+          PACMAN="$1"
+        else
+          gettext "Missing --pacman argument"
+          echo
+          gettext "$(usage)"
+          exit 1
+        fi
         ;;
       --pacman-conf)
-        [[ -n "$2" ]] || (gettext "Missing --pacman-conf argument" && echo && gettext "$(usage)" && exit 1)
-        shift
-        PACMAN_CONF="$1"
+        if [[ -n "$2" ]]; then
+          shift
+          PACMAN_CONF="$1"
+        else
+          gettext "Missing --pacman-conf argument"
+          echo
+          gettext "$(usage)"
+          exit 1
+        fi
         ;;
       --arch)
-        [[ -n "$2" ]] || (gettext "Missing --arch argument" && echo && gettext "$(usage)" && exit 1)
-        shift
-        DOWNGRADE_ARCH="$1"
+        if [[ -n "$2" ]]; then
+          shift
+          DOWNGRADE_ARCH="$1"
+        else
+          gettext "Missing --arch argument"
+          echo
+          gettext "$(usage)"
+          exit 1
+        fi
         ;;
       --ala-url)
-        [[ -n "$2" ]] || (gettext "Missing --ala-url argument" && echo && gettext "$(usage)" && exit 1)
-        shift
+        if [[ -n "$2" ]]; then
+          shift
         DOWNGRADE_ALA_URL="$1"
+        else
+          gettext "Missing --ala-url argument"
+          echo
+          gettext "$(usage)"
+          exit 1
+        fi
         ;;
       --ala)
         DOWNGRADE_FROM_ALA=1
@@ -313,11 +337,9 @@ cli() {
   done
 
   if ((!"${#terms[@]}")); then
-    (
-      gettext "No packages provided for downgrading"
-      echo
-      gettext "$(usage)"
-    ) >&2
+    gettext "No packages provided for downgrading"
+    echo
+    gettext "$(usage)"
     exit 1
   fi
 
--- a/downgrade
+++ b/downgrade
@@ -326,8 +326,8 @@ cli() {
         gettext "$(usage)"
         exit 1
         ;;
-      --)
-        ;;
+      --) ;;
+
       -*)
         pacman_args=("$@")
         break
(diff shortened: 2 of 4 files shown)


downgrade
```diff
@@ -272,7 +353,9 @@ cli() {
     }
   fi
 
-  main "${terms[@]}"
+  for term in "${terms[@]}"; do
+    main "$term" || (gettext "No package with name $term found" && echo && gettext "$(usage)" && exit 1)
+  done
```
pbrisbin, line 358: I'd rather not print `usage` on this error. It's not invalid usage IMO, so the help text would just be noise that could obscure the real error.

Also, this is a behavior change. Currently, `downgrade` will try all arguments, and exit `1` if any fail. This loop now exits on any failure immediately. I'm not necessarily against that, but you're going to want to remove the failure-handling in `main` if that's the behavior you're after.

Personally, I like it the way it is, and you could add a `gettext` error message within the `main` loop, if you just want some more user feedback.

downgrade
```diff
@@ -243,25 +267,82 @@ declare -a installed
...
+        DOWNGRADE_FROM_CACHE=1
+        ;;
+      --nosudo)
+        DOWNGRADE_NOSUDO=1
+        ;;
+      -h | --help)
+        gettext "$(usage)"
+        exit 1
+        ;;
+      --) ;;
```
pbrisbin, line 329: This branch should (now) `shift`, read `$@` into `terms`, and `break`. That way, in the (super duper unlikely) event of a package name that begins with a hyphen, you can use `--` to be able to pass it as an argument, instead of it being swallowed in `pacman_args`.

downgrade
```diff
@@ -4,6 +4,30 @@
...
+    --pacman <command>
+                    pacman command to use, defaults to "pacman"
+    --pacman-conf <directory>
+                    pacman configuration file, defaults to "/etc/pacman.conf"
+    --arch <architecture>
+                    target architecture, defaults to output of "uname -m"
+    --ala-url <url>
+                    location of ALA server, defaults to "https://archive.archlinux.org"
+    --ala           only use ALA server
+    --cached        only use cached packages
```
pbrisbin, line 21: I think it's ambiguous in the option from that `--x` means _only_ `x`. WDYT of either:

- Re-inverting these to `--no-x` to not use `x`, or
- Naming these `--only-x` or `--x-only`?

doc/.gitignore
```diff
@@ -0,0 +1,2 @@
+todos.org
+Peek*
```
pbrisbin, line 2: What's all this?

atreyasha: My bad, removed these.

downgrade
atreyasha: Agree, changed them to variant `--x-only`

downgrade
atreyasha: Yes, that would be a good failsafe in that contingency. To keep the code as similar to the latest version of downgrade (before this PR), I reverted the branch to `shift` and `break` for the `pacman_args` instead of `terms`. I also updated the `usage` script to describe the syntax with `--`.

downgrade
atreyasha: Agree with that. I added the `gettext` for user feedback when the package name supplied does not exist. This is now directly in `main`. However, this leads to an error with in tests `test/main/invalid-choice.t` and `test/main/no-candidates.t` because this prints text where it was not previously expected. Should the test script be modified for this case?

downgrade
```diff
@@ -4,6 +4,30 @@
 export TEXTDOMAIN=downgrade
 export TEXTDOMAINDIR=/usr/share/locale
 
+usage() {
+  cat <<EOF
+Usage: downgrade [options] [packages] -- [pacman_options]
```
pbrisbin, line 9: I'm not 100% expert in usage syntax, but I think this should be:

```suggestion
Usage: downgrade [options] [packages] [-- [pacman_options]]
```

Thank you for considering compatibility, and it's nice that hyphen-prefixed package names still work.

I'm willing to change things and bump the major version if you think the usage with `[--] [packages]` is nicer. We can ship this one as-is with a minor version bump, then switch that later, or do it as part of this. Up to you.

downgrade
```diff
@@ -243,25 +268,81 @@ declare -a installed
...
   if ((!"${#terms[@]}")); then
-    (
-      gettext 'usage: downgrade <pkg> [pkg...] [-- pacman_option...]'
-      echo
-      gettext 'see downgrade(8) for details.'
-      echo
-    ) >&2
+    gettext "No packages provided for downgrading"
+    echo
+    gettext "$(usage)"
```
pbrisbin, line 345: This is a behavior change, you're now outputting these errors to `stdout`, not `stderr` like before.

downgrade
```diff
@@ -209,6 +233,7 @@ main() {
     # shellcheck disable=SC2207
     installed=($(previously_installed "$term"))
     current=$(currently_installed "$term")
+    [[ -z "$current" ]] && gettext "No package with name $term found"
     process_term "$term" || exit_code=1
```
pbrisbin, line 237: This isn't _exactly_ right. We now fail if we're the term isn't currently installed, but there's actually no reason to fail there. If I want to install some older `firefox`, even if I don't currently have `firefox` installed, I think that should work -- and it used to.

I think you want:

```suggestion
    if ! process_term "$term"; then
      gettext "Unable to downgrade $term"
      exit_code=1
    fi
```

downgrade
```diff
@@ -187,7 +211,7 @@ process_term() {
     choice=${candidates[0]}
   else
     if present_packages "$term" "${candidates[@]}"; then
-      choice=$(read_selection "${candidates[@]}")
+      choice=$(read_selection "${candidates[@]}") || gettext "Invalid choice"
```
pbrisbin, line 214: Did you test this behavior? I'm not sure `read_selection` exits non-zero for an invalid choice, or it just prints nothing. You might want to check for `[ -z "$choice" ]` instead. Also, should this error, or just print and continue?

downgrade
atreyasha: Yes I overlooked that possibility. Your suggestion will keep that functionality going, committed.

downgrade
atreyasha: Not an expert on syntax either, but yes the original usage was misleading. I found a nice summary for usage syntax here: http://docopt.org/

Will modify the usage and commit again.

downgrade
atreyasha: Yes true. I will modify error messages to go explicitly to stderr. Sorry, am a bit rusty with my bash :smile:

Committed the change.

downgrade
```diff
@@ -188,6 +213,9 @@ process_term() {
   else
     if present_packages "$term" "${candidates[@]}"; then
       choice=$(read_selection "${candidates[@]}")
+      if [ -z "$choice" ]; then
+        ( gettext "Invalid choice" && echo ) >&2
```
pbrisbin, line 217: I don't think you need these parenthesis. Ditto on L241. (The ones at L295 for example are needed because you're grouping multiple expressions, though `{ ... }` may be better there.)

downgrade
atreyasha: It prints nothing when there is an invalid choice. Modified `process_term` to check `choice` and output to stderr. Now, it will print `Invalid choice` and will continue back to `main` where it will print another error message `Unable to downgrade foo`. I think it is fine this way.

downgrade
```diff
@@ -243,24 +274,90 @@ declare -a installed
...
         terms+=("$1")
-        shift
         ;;
     esac
+    shift
   done
 
   if ((!"${#terms[@]}")); then
     (
-      gettext 'usage: downgrade <pkg> [pkg...] [-- pacman_option...]'
```
pbrisbin, line 260: Oh, actually, I think this usage was following a defined convention. See https://github.com/pbrisbin/downgrade/issues/69.

I think you want your new usage to be:

```
usage: downgrade [option...] <pkg> [pkg...] [-- pacman_option...]
```

atreyasha: Committed a correction

downgrade
atreyasha: Committed a fix

atreyasha:
Two more comments:

1. I modified `tests/main/invalid-choice.t` and `tests/main/no-candidates.t` to send stderr to `/dev/null` since the error messages were messing with the tests. Now they all succeed.

2. Regarding translation, yes do add me as a contributor. I am working and studying in machine translation, so this is something I should (theoretically) be able to manage :smile:

downgrade
atreyasha: Also added two possible syntaxes in the usage, since both are technically possible. Also I see many other CLI tools using alternate syntaxes which I personally find cool.

```
Usage: downgrade [option...] <pkg> [pkg...] [-- pacman_option...]
       downgrade <pkg> [pkg...] [option...] [-- pacman_option...]
```

downgrade
```diff
@@ -4,6 +4,31 @@
 export TEXTDOMAIN=downgrade
 export TEXTDOMAINDIR=/usr/share/locale
 
+usage() {
+  cat <<EOF
+Usage: downgrade [option...] <pkg> [pkg...] [-- pacman_option...]
+       downgrade <pkg> [pkg...] [option...] [-- pacman_option...]
```
pbrisbin: I don't think we need both lines. If you're trying to indicate that you can interleave options with packages, this doesn't even convey that anyway (it just shows "all packages, all options" or "all options, all packages"). I think we can just show the first line and let the user assume interleave-ability.

atreyasha: Alright, fixed.

pbrisbin:
> Regarding translation, yes do add me as a contributor

I think you have to sign up at https://gitlocalize.com/, then let me know your username. I tried just adding your GitHub handle and it said "invalid username" :shrug: 

> I could even make an automated pipeline so it would translate all tokens which are called with gettext

I've had a lot of native-speaker contributors over the years for translations, so I might want to avoid pushing that off on automation just yet. Ideally, I'd be able to reach out to those previous maintainers when we add new keys, then resort to machine translations after some time, if there's no response. I'd love it if GitLocalize helped manage that contributor workflow, but the project is so new I'm just not sure yet.

> I think you would probably only want to ship this to the AUR after we have the translations settled

That's right, but I'm not against merging and then not releasing (yet). It's a slight risk in case some bugfix comes along and master is not releasable, so I probably wouldn't prefer this option.

> you could merge these changes to another branch and you could add me as a collaborator, and I can work on that branch till I finish the translations

I'm fine with that, if it's what you want. Just curious: how would that be different than just keeping this branch open? Or you branching from this branch yourself to isolate the translation work?

I should've though to make you a collaborator anyway, so you didn't have to work via forks all the time. I'll go ahead and do that, whatever else we decide.

atreyasha:
> I think you have to sign up at https://gitlocalize.com/, then let me know your username. I tried just adding your GitHub handle and it said "invalid username" 

Done, my username is `atreyasha`. Let me know what I should do.

> That's right, but I'm not against merging and then not releasing (yet). It's a slight risk in case some bugfix comes along and master is not releasable, so I probably wouldn't prefer this option.

Agreed, is a smarter option to keep the changes in a non-master branch till we are ready to release.

> I'm fine with that, if it's what you want. Just curious: how would that be different than just keeping this branch open? Or you branching from this branch yourself to isolate the translation work?

Come to think about it, the difference is only symbolic. Just feels a bit more empowering to contribute on an actual branch than a fork :smile:

Also, this PR is getting long with many messages; which is why I would like to merge it into a non-master branch in your repo, then create a new PR just for the readme, documentation and translation.

pbrisbin:
> Done, my username is atreyasha. Let me know what I should do.

Added. For now, just have a look around the UI and see what you think: https://gitlocalize.com/repo/4232. I know that once we merge un-translated keys those languages would show as no longer 100%, and you can click in and add translations, then open a PR to commit them to the repo. However, if we're going to keep the new keys out of master for now, I'm not sure if there'd be a way to work on it through that UI. Your guess is as good as mine.

> feels a bit more empowering to contribute on an actual branch than a fork smile
>
> Also, this PR is getting long with many messages; which is why I would like to merge it into a non-master branch in your repo, then create a new PR just for the readme, documentation and translation.

If you are done working in this branch, I will squash-merge everything into a new (non-fork) branch. You're a contributor now, so you can go ahead and work as you'd like from there. Just let me know if it's safe to do that now, or not yet.

atreyasha:
> Added. For now, just have a look around the UI and see what you think: https://gitlocalize.com/repo/4232. I know that once we merge un-translated keys those languages would show as no longer 100%, and you can click in and add translations, then open a PR to commit them to the repo. However, if we're going to keep the new keys out of master for now, I'm not sure if there'd be a way to work on it through that UI. Your guess is as good as mine.

I see. Alright I will start exploring and will get back to you later.

> If you are done working in this branch, I will squash-merge everything into a new (non-fork) branch. You're a contributor now, so you can go ahead and work as you'd like from there. Just let me know if it's safe to do that now, or not yet.

Yes I am done with this PR for now, so go ahead with the merge. Thanks again.

PR #102, opened 2020-04-16
CLI options

pbrisbin:
A squashed, non-fork version of #96.


atreyasha:
@pbrisbin thanks for that. Btw, could you open another project in `gitlocalize` and have it track the `as/cli-arguments` branch? I think that would be the best way for me to edit locally in this branch.

pbrisbin:
> Btw, could you open another project in gitlocalize and have it track the as/cli-arguments branch?

Unfortunately, it won't let me pick `downgrade` again for a new project, since it already is one.

atreyasha:
Hi @pbrisbin,

Would just like to summarize the changes I made so far and get your feedback:

1. I modified `README.md` with respective command line option changes and pasted the usage script as I thought this would be more descriptive. I changed the circle CLI badge to the shield-style as IMO it looks better.

2. Correspondingly modified the man-page documentation with bold options, and added myself as an author.

3. Compared to the previous PR, I modified `downgrade` slightly as well. I moved `main`, `pacman` and `prompt_to_ignore` and put them outside the `cli` function. I needed to do this as I added a test script `options_sanity.t` to check that all options are passed correctly. For this, I needed the `cli` function to behave like a parser. I checked and this does not change any behaviour otherwise, so this should be fine.

4. I modified the `zsh` completion script so it autocompletes both packages and command-line options for `downgrade`. After the cursor passes a ` -- ` character sequence, no autocompletions are supplied as it is assumed that `pacman` arguments should follow. I did not want to autocomplete `pacman` arguments as this might get messy.

5. I added a `.gitattributes` file which will no longer counts test `*.t` files as `perl` files and will show the distribution of code to be mostly `shell` and `Makefile` instead of `perl`; which I believe is a more accurate representation of the functional code in this repository. The changes in code statistics will only be reflected once we merge into `master`. This is based off documentation [here](https://github.com/github/linguist).

atreyasha:
If all of the above points are fine, we finally come to the beast of translation.

Based on the files present in `locale`, most of them are unchanged. We have some additional one-liners such as "Invalid choice", which I can easily translate.

1. My first issue is what I should do with the usage script. It is quite long and is not a one-liner `gettext` command like the rest. So should usage entry in the `*.pot` look like:

```
#: downgrade:9
msgid "Usage: downgrade [option...] <pkg> [pkg...] [-- pacman_option...]"
msgstr ""

#: downgrade:11
msgid "  Options:"
msgstr ""

. . .
```

I am bit unsure of how to go about this usage part. Everything else should be fine.

2. Since the `gitlocalize` project is based off the `master` branch, does it make more sense that we merge this branch first (if everything else if okay), and then I work on the translations via `gitlocalize`?

.gitattributes
```diff
@@ -0,0 +1 @@
+test/* linguist-documentation
```
pbrisbin, line 1: Calling it `linguist-documentation` doesn't seem right. Is there a better (not perl) option?

README.md
```diff
@@ -10,9 +10,31 @@ Install via the [AUR](https://aur.archlinux.org/packages/downgrade/).
 
 ## Usage
 
-Downgrade some packages, checking both local cache and the [A.L.A.][ala]:
+Below is the usage script for `downgrade`:
```
pbrisbin, line 13: I would just ditch this sentence now, FWIW.

README.md
```diff
@@ -10,9 +10,31 @@ Install via the [AUR](https://aur.archlinux.org/packages/downgrade/).
...
+    --cached-only   only use cached packages
+    --nosudo        do not use sudo even when available, use su
+    -h, --help      show help script
+
+  Note:
+    Any options after the -- character sequence will be treated as pacman options.
+    See downgrade(8) for further details.
+```
+
+Downgrade packages checking both local cache and the [A.L.A.](https://wiki.archlinux.org/index.php/Arch_Linux_Archive), which is the default behavior.
```
pbrisbin, line 37: :art: can you hard-warp at 80 columns throughout all the markdown?

We should enable `prettier-markdown` in `.restyled.yaml`.

README.md
```diff
@@ -21,20 +43,20 @@ $ downgrade foo bar
...
 ```
-$ DOWNGRADE_FROM_ALA=0 downgrade foo
+$ downgrade --cached-only foo
 ```
 
 Downgrade a package, looking in only the A.L.A.:
 
 ```
-$ DOWNGRADE_FROM_CACHE=0 downgrade foo
+$ downgrade --ala-only foo
```
pbrisbin, line 52: :art: trailing whitespace.

doc/downgrade.8.md
```diff
@@ -1,18 +1,18 @@
-% DOWNGRADE(8) User Manuals
-% Patrick Brisbin <pbrisbin@gmail.com>
-% March 2019
+% DOWNGRADE(8) User Manual
+%
+% April 2019
```
pbrisbin, line 3: I think it's 2020!

downgrade
```diff
@@ -209,7 +239,13 @@ main() {
     # shellcheck disable=SC2207
     installed=($(previously_installed "$term"))
     current=$(currently_installed "$term")
-    process_term "$term" || exit_code=1
+    if ! process_term "$term"; then
+      {
+        gettext "Unable to downgrade $term"
```
pbrisbin, line 244: I think you have to do something here so `$term` doesn't end up in the translation key and never matching. Is this what `eval_gettext` is for?

downgrade
```diff
@@ -271,13 +373,12 @@ cli() {
...
   fi
-
-  main "${terms[@]}"
-  sudo pacman -U "${pacman_args[@]}" "${to_install[@]}"
-  prompt_to_ignore "${to_ignore[@]}"
 }
 
 if ((!LIB)); then
   set -e
   cli "$@"
```
pbrisbin, line 380: I would rename this to `parse_options` or something like that now. Or maybe even do this:

```sh
parse_options() {
  # old contents of most of cli()
}

cli() {
  parse_options "$@"
  main "${terms[@]}"
  sudo pacman -U "${pacman_args[@]}" "${to_install[@]}"
}
  
if ((!LIB)); then
  set -e
  cli "$@"  # unchanged now
fi
```

test/options_sanity/options_sanity.t
```diff
@@ -0,0 +1,47 @@
+  $ source "$TESTDIR/../helper.sh"
+
+Checking that CLI options match up with environmental variables, packages and pacman options
```
pbrisbin, line 3: :clap: love to see tests on new code!

atreyasha:
Thank you for your words, really appreciate the support and the excitement as well :)

> I think so. As long as there aren't unknowns or risk of a big delay, I don't mind merging to master and then solving that relatively quickly before the next release.

> Fair warning: The last (and only) time I tried to add translations via GitLocalize and open the PR to add them in the repo, it didn't work. I reported it to the project maintainers and they said they'd be working on it. I'm hoping it'll be fixed by the time we need it for these, but we'll see!

Hmm, in that case I think it is a better option to just translate it locally. I will try out your suggestions and see how far I get. We can merge it to `master` once the translations are done in this branch.

README.md
atreyasha: Yes agree, just a re-statement.

.gitattributes
atreyasha: `github-linguist` only offers a few options for manually overriding misinterpreted languages. I will change this in a commit later to `*.t -linguist-detectable`, which will make the `*.t` files undetectable by linguist.

doc/downgrade.8.md
atreyasha: Haha wishful thinking maybe

.gitattributes
pbrisbin: Ohhhh, that's my fault. I thought you were saying this was "documentation for linguists", like it was somehow related to translations. Now I see linguist is the library and you're calling these files documentation. That'd be fine with me, as is your other option. Totally up to youl

downgrade
atreyasha: Yes this looks cleaner

README.md
atreyasha: Updated `.restyled.yaml`, the hard-wrapping done too.

downgrade
atreyasha: Ah yes true, that is exactly how `eval_gettext` was used for another option.

atreyasha:
Updated the translations. Essentially copied the old ones for the same text and used Google Translate for the new text bits. They all look good to me for now. In the future, I will work on a script that can update all the line numbers in the translations if `downgrade` is modified.

What should I do with the metadata in each `*.po` file? Should I leave the previous translator's details or write that this is a mix of machine and human translation?

pbrisbin:
>  Should I leave the previous translator's details or write that this is a mix of machine and human translation?

Hmm, good question, I don't think I have an opinion. If the file format supports listing multiple maintainers, it would make sense to me to keep the original name and add yourself (even though you did it via machine). WDYT?

.restyled.yaml
```diff
@@ -3,3 +3,7 @@
     include:
       - downgrade
     image: restyled/restyler-shfmt:v3.0.1
+    
+- prettier-markdown:
+    include:
+      - README.md
```
pbrisbin, line 9: :+1:

How bad is it if you let prettier-markdown run on the manpage source? If it makes a crazy diff, we can hold off for now, but if it's not too bad I think it'd be good to wrap Restyled on that too, or any markdown in the repo (e.g. the default `include`).

atreyasha: Let's give it a try, committing it now.

atreyasha: Hmm, for the manpage source markdown it seems to interfere with some of the ultimate manpage output. I would prefer if we exclude this from restyled since this manpage markdown is not exactly something we present but is more of a tool for us to generate the manpage.

atreyasha:
> Hmm, good question, I don't think I have an opinion. If the file format supports listing multiple maintainers, it would make sense to me to keep the original name and add yourself (even though you did it via machine). WDYT?

Makes sense, will just do that now.

> This LGTM! Let me know when you're ready and I'll squash-merge and release.
> 
> This can still be a minor version bump only?

Minor version bump sounds good. Get back to you in a bit.

atreyasha:
Done with the updates to the `*.po` files. If all is well, we can squash merge.

Once we merge, I would just wait a few minutes to see if the GitHub language statistics gets updated. If not, I might try to update `.gitattributes` directly in master to see if it works. Hope that would be fine.

pbrisbin:
> I might try to update .gitattributes directly in master to see if it works. Hope that would be fine

SGTM.

> maybe you could do a local install of this branch and see if the zsh autocompletions work for you as well (I think you use zsh)

Seems to work well to me! The only thing I noticed is that after giving `--arch`, it is completing package names, where ideally it would complete between the two architecture options -- but that's not a big enough deal to block merge. Maybe file as a low-priority issue?

atreyasha:
> > I might try to update .gitattributes directly in master to see if it works. Hope that would be fine

Language stats worked!

> Seems to work well to me! The only thing I noticed is that after giving `--arch`, it is completing package names, where ideally it would complete between the two architecture options -- but that's not a big enough deal to block merge. Maybe file as a low-priority issue?

Will do that. Happy for the merge :)

Issue #82, opened 2019-09-15
Downgrade via version number on command line


Gibbz:
Id like to be able to use downgrade to work with a supplied version number on the commandline.
For example i want to downgrade wine to version 4.9 which downgrade shows in the second column.
For example:
downgrade wine-4.9

And also an option like 'pacaur --noconfirm --needed' which will auto install the package

pbrisbin:
> Id like to be able to use downgrade to work with a supplied version number on the commandline

Downgrade has the following two behaviors:

- If you pass `{package}-{version}`, it will limit the results shown
- If you limit the results shown to a single package, it will proceed without prompting

Therefore, as long as you supply something specific enough, it will just downgrade straight away.

Your example seems to work:

```console
% downgrade wine-4.9
wine-4.9-1-x86_64 27.3 MiB  14.3M/s 00:01 [##############################-----------------------]  55
```

> And also an option like 'pacaur --noconfirm --needed' which will auto install the package

Are you talking specifically about `--noconfirm`? I'm not sure if `--needed` would be useful.

Either way, making `downgrade` respect environment variables such as `PACMAN` or `PACMAN_OPTIONS` (or both) would be one way to do this. I would except such a PR, but I'm unlikely to have time to get to it myself.

If I understand the request correctly (please confirm), I'll create an isolated Issue for it and add the appropriate tags. Let me know!

winiciuscota:
Is there anyway to specify whether to use local or remote?

pbrisbin:
Yup, either the searching of the remote archive or the local cache can be disabled by env var, check `man downgrade`.

pbrisbin:
Closing as stale.

It sounds like the Issue as reported is not an Issue, but the following feature did come up:

> > And also an option like 'pacaur --noconfirm --needed' which will auto install the package
>
> Are you talking specifically about --noconfirm? I'm not sure if --needed would be useful.
> 
> ... making downgrade respect environment variables such as PACMAN or PACMAN_OPTIONS (or both) would be one way to do this.

Anyone interested should feel free to create a separate Issue with details on this feature request.

atreyasha:
@Gibbz

>Id like to be able to use downgrade to work with a supplied version number on the commandline.

With #117, we now have version filtering operators which can perform such operations. See the readme for some examples.

>And also an option like 'pacaur --noconfirm --needed' which will auto install the package

After merging #118, we should be able to implement auto-installation (without further prompts).

@winiciuscota 

[...]

PR #108, opened 2020-05-01
Enhanced version-bounding logic and related changes

atreyasha:
Hi @pbrisbin, I will just enumerate the changes in this branch and also ask some questions.

### Changes

1. Version bounds have been implemented. Now, the user can input version bounds, for example `downgrade vim=8.1` or `downgrade "vim>8.1"`. Supported logical operators are `=|<|>|<=|>=`.

    The previous functionality where a user could input a `-` (eg. `downgrade vim-8.1`) to specify package regex has now been replaced with `=` (eg. `downgrade vim=8.1`) as this is IMO more intuitive. 

    The proposed logical operators comply with pacman's version bounding standards, which will help us in the next step of handling version-bound dependencies. Furthermore, this change addresses issue #82. Package version comparison is done with the aid of `vercmp` (https://www.archlinux.org/pacman/vercmp.8.html), which is a `pacman` tool.

    **Qn:** `Restyled` is throwing errors on line [267](https://github.com/pbrisbin/downgrade/blob/as/version-logic/downgrade#L267) of `downgrade`, it is not able to interpret the regex. Any suggestion on what I could do?

2. Tests have been updated to test the version bounding. However, since our Circle CLI checker is built on a Debian-based system, it cannot run the arch-based `vercmp` and shows a failing status.

    **Qn:** Is there any way to change the Docker image of our Circle CLI instance to an arch image? I think we could use a public Docker Hub archlinux image, see (https://circleci.com/docs/2.0/executor-types/#public-images-on-docker-hub) and (https://hub.docker.com/_/archlinux?tab=description).

3. Logic for the `--arch` command-line option has been removed and the documentation has been updated to reflect this. I also removed the test corresponding to finding packages with the `i686` architecture since this is now redundant.
 
4. There was a small bug in the ignore-package behaviour, in that if one executed `downgrade vim-8.2.0510-1` (eg. with old syntax), then at the end the package to be ignored will be outputted as `vim-8.2.0510.1`. I modified this such that only the real package name `vim` will be sent for ignoring. 

5. I updated our `Makefile` and made some of the locale commands simpler. Also, I modified the `locale` command to firstly create a new `downgrade.pot` template and then update all existing locales with this template to reflect line number changes and to comment out redundant chunks (for example the `--arch` chunks being commented out in this iteration).

    Here, I have another question. I also want to add a `Makefile` segment which installs all locales to my system so I can check that they work fine. Ideally, I would like this to be part of the `Makefile` to make it easier for testing/development. Following is the shell-script command based on the AUR package-build:

    ```shell
    #!/bin/bash
    # assume pwd is ./locale/

    for po_file in ./*.po; do
        locale="$(basename "$po_file" .po)"
        mkdir -p "/usr/share/locale/$locale/LC_MESSAGES/"
        msgfmt "$po_file" -o "/usr/share/locale/$locale/LC_MESSAGES/downgrade.mo"
    done
    ```

    **Qn:** Any recommendations on how to include this in the `Makefile`? Would it be okay if I made a separate shell-script with this code labelled eg. `locale.sh` and executed that via the `Makefile`?

6. I added additional GitHub badges (see this branch's readme) to reflect GitHub and AUR version numbers, just thought it looks nice. But maybe you can provide a second opinion if this is overkill :smile:

### Next steps

Let me what you think of these changes. If all is fine, I will update the readme, man-page documentation and changelog with the updates.


pbrisbin:
> Qn: Restyled is throwing errors on line 267 of downgrade, it is not able to interpret the regex. Any suggestion on what I could do?

That seems like a bug in `shfmt` that you could report. It's a shame to have to workaround this deficiency, but I'd rather not ignore the whole `downgrade` file (e.g. the whole project) from Restyled, so I think it's worth trying a few things:

One option is to try extracting a variable,

```sh
regex='(...|...|...|)'

if [[ "$temp" =~ $regex ]]; then
```

Another option is to just not have the conditional,

```sh
    breakdown=($(sed -r "s/(.*)\b(>=|<=|<|>|=)\b(.*)/\1 \2 \3/g" <<< "$term"))
    term="${breakdown[0]}"
    operator="${breakdown[1]}"
    version="${breakdown[2]}"
```

Can the `sed` be written such that, if the pattern doesn't match, you end up with all of `term` in `breakdown[0]` and empty `breakdown[1,2]`? If so, it would function the same way and not need the conditional at all.

> Qn: Is there any way to change the Docker image of our Circle CLI instance to an arch image? I think we could use a public Docker Hub archlinux image, see (https://circleci.com/docs/2.0/executor-types/#public-images-on-docker-hub) and (https://hub.docker.com/_/archlinux?tab=description).

Yes, running CI in Arch would be great. Feel free to experiment with whatever Docker image you think may work.

Historically, I would use [mocking](https://pbrisbin.com/posts/mocking_bash/) to fake any Arch-specific commands in tests. But since you are actually testing `vercmp`, I don't know that mocking it is a great idea. That said, why are we testing how `vercmp` behaves? Maybe mocking it _is_ a good idea?

> Qn: Any recommendations on how to include this in the Makefile? Would it be okay if I made a separate shell-script with this code labelled eg. locale.sh and executed that via the Makefile?

If this routine is useful to have in the project `Makefile`, that sounds good to me. I can update the `PKGBUILD` to call the make task too.

I'm not against extracting a separate script when it makes sense, but this routine is relatively small, so I would probably just in-line it:

```make
.PHONY: install.po
install.po:
        for po_file in ./*.po; do \
          locale="$$(basename "$$po_file" .po)" && \
          mkdir -p "/usr/share/locale/$$locale/LC_MESSAGES/" && \
          msgfmt "$po_file" -o "/usr/share/locale/$$locale/LC_MESSAGES/downgrade.mo"; \
        done
```

Makefile
```diff
@@ -1,3 +1,4 @@
+LOCALE_PREFIX ?= /usr
```
pbrisbin, line 1: Does `LOCALE_PREFIX=$(PREFIX)/share/locale` not work? I don't like how this is inconsistent with to `MANPREFIX`

CHANGELOG.md
```diff
@@ -1,11 +1,18 @@
-## [*Ongoing*](https://github.com/pbrisbin/downgrade/compare/v6.3.0...master)
+## [*Ongoing](https://github.com/pbrisbin/downgrade/compare/v6.4.0...master)
```
pbrisbin, line 1: You've dropped the second `*` on this heading and the next, breaking the bolding.

Makefile
```diff
@@ -7,18 +8,17 @@ setup:
 	command -v vbump  || aurget vbump-git
 	command -v pandoc || stack install pandoc
 
-locale/downgrade.pot: downgrade
+locale:
```
pbrisbin, line 11: Please leave these targets how they were. They ensure that the rule is not run if the output file is up to date from the dependencies. What you've created here would be a "phony" though you've not annotated it as such. These are fine to have in addition (see `man`), but not instead of.

atreyasha: I had added `locale` as a "phony" below, but I see the point keeping it this way. Adding it back.

CHANGELOG.md
atreyasha: My bad. I will just remove the bolding since most of the older entries are not bold, and this is anyways a header.

Makefile
atreyasha: The reason I had to make this different is that the locales have to installed at the `/usr/share/locale` prefix instead of the `/usr/local/share/locale` prefix which would result from `$(PREFIX)/share/locale` (which maps to `MANPREFIX`). This is predefined because of `export TEXTDOMAINDIR=/usr/share/locale` in `downgrade`.

I mofieid `LOCALE_PREFIX` to be slightly more similar to `MANPREFIX`.

atreyasha:
> Can the `sed` be written such that, if the pattern doesn't match, you end up with all of `term` in `breakdown[0]` and empty `breakdown[1,2]`? If so, it would function the same way and not need the conditional at all.

This one worked and is the simplest solution.

> Yes, running CI in Arch would be great. Feel free to experiment with whatever Docker image you think may work.
> 
> Historically, I would use [mocking](https://pbrisbin.com/posts/mocking_bash/) to fake any Arch-specific commands in tests. But since you are actually testing `vercmp`, I don't know that mocking it is a great idea. That said, why are we testing how `vercmp` behaves? Maybe mocking it _is_ a good idea?

Working on this, will need some time to figure it out. We are not exactly testing `vercmp` but more on just checking that the function calling `vercmp` is not buggy when dealing with the multiple regular expressions.

> If this routine is useful to have in the project `Makefile`, that sounds good to me. I can update the `PKGBUILD` to call the make task too.

Added "phonies" for `install.po` and `uninstall.po`

Makefile
atreyasha: WDYT?

Makefile
atreyasha: I understand. Added the original behaviour back, but also added `locale` as a phony so it is easier to call. I also fixed the workflow for the segment with `downgrade.8`, such that the `man` doc is only made if there are changes.

atreyasha: The phony annotations are collated on line [41](https://github.com/pbrisbin/downgrade/blob/f330159a4813031a0209628650cade4740012873/Makefile#L41)

atreyasha:
Update: Managed to make the circle CI test work. I found a python package which performs the same function as `vercmp`, [here](https://github.com/sassoftware/python-rpm-vercmp). Then I just mocked `pacman`'s `vercmp` as this pythonic one, and it seems to work fine.

I will just add a few more tests and start updating the documentation.

Makefile
pbrisbin: > This is predefined because of export `TEXTDOMAINDIR=/usr/share/locale` in downgrade.

Ah crap, that's a pre-existing packaging bug then. If one were to `make install PREFIX=/opt`, _everything_ should live in `/opt`, and `downgrade` should use `TEXTDOMAINDIR=/opt/share/locale`. So really, my `Makefile` suggestion is correct, but the hard-coded `TEXTDOMAINDIR` is not right.

I see three options:

- Ignore this (for now), and keep what you're doing
- Do as I suggested for `LOCALE_PREFIX`, and make it work by:

  Making `downgrade` a `.in` file, with an interpolation for `@PREFIX@` in the `TEXTDOMAINDIR` definition. Pre-process `downgrade.in` to `downgrade`, filling in `PREFIX` along the way.

- Do as I suggested for `LOCALE_PREFIX`, and make it work by:

  Writing a file as part of `make install`, such as `/etc/profile.d/dowgrade.sh`, which includes the `export TEXTDOMAINDIR` with a `PREFIX`-prefixed value. `souce` this `.env` file from the `downgrade` script.

Do you see any other options? What do you think of those 3?

Makefile
pbrisbin: > The phony annotations are collated on line

Oh, my bad. I forgot I used to do this style. Lately, I've liked putting the `.PHONY` right above the target it's annotating.

pbrisbin:
> We are not exactly testing vercmp but more on just checking that the function calling vercmp is not buggy when dealing with the multiple regular expressions
>
> I found a python package ... Then I just mocked

I haven't looked at the diff yet, but this is a little confusing. If you're going to use mocking of `vercmp` (via `test/bin/act-like`), that's all you need to do: create `test/bin/vercmp`, run the test suite locally (on Arch) and you're done. CI would use recorded responses, it wouldn't need any real `vercmp`, python or otherwise.

CHANGELOG.md
```diff
@@ -1,13 +1,21 @@
...
+- [FIX] Modified Makefile to handle more thorough locale updates
+
+## [v6.3.0](https://github.com/pbrisbin/downgrade/compare/v6.1.0...v6.3.0)
 
 - [NEW] Replace environmental variables with command line arguments/options
-- [NEW] Updating locales to reflect new keys
+- [FIX] Updating locales to reflect new keys
 
-## [*v6.1.0*](https://github.com/pbrisbin/downgrade/compare/v6.0.0...v6.1.0)
+## [v6.1.0](https://github.com/pbrisbin/downgrade/compare/v6.0.0...v6.1.0)
```
pbrisbin, line 18: :+1:

Makefile
```diff
@@ -1,5 +1,6 @@
 PREFIX    ?= /usr/local
 MANPREFIX ?= $(PREFIX)/share/man
+LOCALE_PREFIX ?= /usr/share/locale
```
pbrisbin, line 3: If we decide not to fix the `TEXTDOMAINDIR` packaging bug, I think maybe we don't use an override-able variable here (overriding it would just break `downgrade`). I would just hard-code the same path again wherever you're using it.

Makefile
```diff
@@ -12,12 +13,15 @@ locale/downgrade.pot: downgrade
 		--from-code=utf-8 -L shell \
 		--package-name=downgrade \
 		--copyright-holder=$(AUTHOR) \
-		-o $@ $<
+		-o locale/downgrade.pot downgrade
```
pbrisbin, line 16: Can you keep using `$@` and `$<` instead of `locale/downgrade.pt` and `downgrade`? This way it'll keep working without any changes if we change these filenames (which is not unrealistic, given `downgrade -> downgrade.in` as a suggestion on the table)

Makefile
```diff
@@ -12,12 +13,15 @@ locale/downgrade.pot: downgrade
...
+	ls ./locale/*po | xargs -I {} msgmerge --update {} ./locale/downgrade.pot
```
pbrisbin, line 17: This now means `locale/*po` should be added as dependencies to this target. If one of those change (but `downgrade` doesn't), we want this rule to fire.

pbrisbin: Can you replace `ls | xargs` with `find -exec`? See [ParsingLs](https://mywiki.wooledge.org/ParsingLs).

Or better, once `locale/*po` are listed pre-requisites, there is an automatic variable for them (`$^`):

```make
locale/downgrade.pot: downgrade locale/*po
        xgettext \
          -o $@ $<
        msgmerge --update $^ $@
```

One bother is that `$^` is all pre-requisites, so it includes `downgrade` in the list. Will that error? I wonder what the nicest way to chop that off would be :thinking:

downgrade
```diff
@@ -105,33 +103,63 @@ prompt_to_ignore() {
...
-    version="${BASH_REMATCH[1]}"
-    name=${1%-$version}
+  if [[ "$operator" = "=" ]]; then
     version="$version.*"
   else
+    non_regex_version="$version"
     version="[^-]+-[0-9.]+"
-    name=$1
   fi
   pkgfile_re="$name-$version-(any|$DOWNGRADE_ARCH)\\.pkg\\.tar\\.(gz|xz|zst)"
```
pbrisbin, line 136: I would offer three broad suggestions, which will (unfortunately) impact a lot of different parts of this diff, so I'll comment it here and submit this review and see what you think before reviewing any more:

**1- Commit to owning filtering the versions yourself, and do so in just one place**

That means, for example, this regex-based `find` now becomes simply

```sh
find $PACMAN_CACHE -maxdepth 1 -name "$name-*.pkg.*"
```

(I don't care about being so specific with the extension.)

**2- Don't break `name-operator-version` as early as you are**

It seems to just require passing 3 variables around and doing repeated conditionals. Instead, I would keep the argument as a single thing as much as possible, and pass it around until it gets to the "one place" I mention in (1). Then (and only then) split it and deal with it accordingly. This will require some version of (3) to work well, I think.

(OK, maybe you need to break out `name` to do the `find` I mention in (1), but I would do that right there where you need it, instead of accepting it pre-broken into this area of the script, personally).

**3- Write `filter_packages` as a, well, filter**

```sh
filter_packages()
  local name_version_filter=$1 pkg

  while read -r pkg; do
    if matches_name_version_filter "$name_version_filter" "$pkg"; then
      echo "$pkg"
    fi
  done
}

matches_name_version_filter() {
  local name_version_filter=$1 pkg=$2

  case name_version_filter in
    *>=*)
      # ...
      ;;
    *<=*)
      # ...
      ;;
    *=*)
      # ...
      ;;
    *)
      # assume name[-version]
    ;;
  esac
}
```

Now, slotting that in here becomes a far nicer read (and smaller diff):

```sh
find $PACMAN_CACHE -maxdepth 1 -name "$name-*.pgk.tar.(gz|xz|zst)" |
  filter_packages "$name_version_filter"
```

Makefile
atreyasha: Yes, reverted it back.

Makefile
atreyasha: Changed all `ls` related pipes to `find` to address this. Also changed all target/dependency inputs to their special tokens.

> This now means `locale/*po` should be added as dependencies to this target. If one of those change (but `downgrade` doesn't), we want this rule to fire.

From my perspective, the typical workflow for updating `*po` files should be as follows:

1. Update `downgrade`
2. Run `make locale`, which will update `downgrade.pot` and apply the template to all `*po` files, update line numbers, comment out old keys, add new keys where necessary
3. Manually add translations for new keys in `*po` files. This should be the last step.

As a result, I think it is part of the process the the `*po` files should be manually updated last and thus it would not make sense to have them part of the dependencies, as they are the final product.

Makefile
atreyasha: I will restyle such that the "phony" annotations are right above the targets, so it is easier to see next time.

Makefile
atreyasha: I will mark this as resolved, feel free to undo it if we should discuss this further.

atreyasha:
> > We are not exactly testing vercmp but more on just checking that the function calling vercmp is not buggy when dealing with the multiple regular expressions
> > I found a python package ... Then I just mocked
> 
> I haven't looked at the diff yet, but this is a little confusing. If you're going to use mocking of `vercmp` (via `test/bin/act-like`), that's all you need to do: create `test/bin/vercmp`, run the test suite locally (on Arch) and you're done. CI would use recorded responses, it wouldn't need any real `vercmp`, python or otherwise.

Aha I understand now. Did not know this was possible. Will simplify as suggested.

Makefile
atreyasha: I understand now. Suggestion 2 with `downgrade.in` seems the least intrusive for me and will probably result in the smallest diff. I will commit a suggestion for this and get your feedback.

Makefile
atreyasha: Will fix this with my suggestion for `downgrade.in`

downgrade
atreyasha: > **1- Commit to owning filtering the versions yourself, and do so in just one place**
> 
> That means, for example, this regex-based `find` now becomes simply
> 
> ```shell
> find $PACMAN_CACHE -maxdepth 1 -name "$name-*.pkg.*"
> ```
>
> (I don't care about being so specific with the extension.)

As far as I understand (and also from experimentation), the currently used explicit regular expression is required in order to prevent fuzzy matches. For example ` $name-*.pkg.*` might result in `vim` matching with `vim-runtime`. I kept it explicit to avoid this.

> **2- Don't break `name-operator-version` as early as you are**
> 
> It seems to just require passing 3 variables around and doing repeated conditionals. Instead, I would keep the argument as a single thing as much as possible, and pass it around until it gets to the "one place" I mention in (1). Then (and only then) split it and deal with it accordingly. This will require some version of (3) to work well, I think.
> 
> (OK, maybe you need to break out `name` to do the `find` I mention in (1), but I would do that right there where you need it, instead of accepting it pre-broken into this area of the script, personally).

Some degree of splitting is already necessary early on in `main` and `process_term` in order to send the clean package names (without operator/version) to `stderr` in case of an error and to send clean package names into `to_ignore`, `previously_installed` and `currently_installed`. I managed to do this less intrusively by working with a single array called `name_op_ver`, which keeps function arguments to single ones.

> 
> **3- Write `filter_packages` as a, well, filter**
> 
> ```shell
> filter_packages()
>   local name_version_filter=$1 pkg
> 
>   while read -r pkg; do
>     if matches_name_version_filter "$name_version_filter" "$pkg"; then
>       echo "$pkg"
>     fi
>   done
> }
> 
> matches_name_version_filter() {
>   local name_version_filter=$1 pkg=$2
> 
>   case name_version_filter in
>     *>=*)
>       # ...
>       ;;
>     *<=*)
>       # ...
>       ;;
>     *=*)
>       # ...
>       ;;
>     *)
>       # assume name[-version]
>     ;;
>   esac
> }
> ```
> 
> Now, slotting that in here becomes a far nicer read (and smaller diff):
> 
> ```shell
> find $PACMAN_CACHE -maxdepth 1 -name "$name-*.pgk.tar.(gz|xz|zst)" |
>   filter_packages "$name_version_filter"
> ```

Thank you for this suggestion. I incorporated this into the code and it looks much cleaner now. Let me know your thoughts.

Makefile
atreyasha: I am not so familiar with `configure` and `make` routines, so I just committed the simplest solution I could think of. Converted `downgrade` to `downgrade.in` with a special variable `@LOCALE_PREFIX@` which needs to be substituted before installation. Updated `Makefile` to reflect this. The installation process for the `/opt` prefix would be:

```
$ PREFIX=/opt make config install install.po
```

atreyasha:
> Unfortunately, the PR contains a lot of different distinct changes of varying size, value, and compatibility, and the git history is quite noisy too. I'd like to rebase/rewrite things to roughly the following distinct commits:

Awesome. Yes, it is a good idea to rebase for different changes.
 
> 1. Minor packaging fixes
>    
>    * Fix extra bolding in CHANGELOG
>    * Add new batches, punctuation in `README`
>    * Move `.PHONY` annotations in `Makefile`
>    * Fix for `doc/downgrade.8` target in `Makefile`
>    * Add the `msgmerge` update and commit fixed `downgrade.pot`
>    * Add `{install,uninstall.po}` targets
> 2. Fix for `TEXTDOMAINDIR`, with `downgrade.in`
>    I have some more `Makefile` nits here. If we get this in its own PR, I'd be happy to just tweak it myself.
> 3. Remove `--arch`
>    Major version bump.
> 4. Implement versioning argument
>    Minor version bump, if `{name}-{version}` still works. Major bump if we're saying we only support `{name}={version}` now.

Agree with these. For 4, the changes would only support `{name}={version}` (or with other operators such as `=|<=|>=|<|>`) and no longer support `{name}-{version}`. So it would need to be a major bump. 

> How is your `git`-foo? Would you be willing to do the history slicing for that, or do you mind if I did?

Sorry about the mess, I am new to software development (and any standards whatsoever) so I really just do things haphazardly and clean them up at the end. I am not very good with history slicing, so it would perhaps be better if you are more familiar. But I feel bad that you would need to seive through all of the commits.

Is there something I could do to simplify this?

pbrisbin:
> Is there something I could do to simplify this?

Not really no, but it's not a big deal. For reference, I wouldn't sift through the commits themselves, I would actually sift through the diff -- which in this case is a bit easier, IMO.

```console
% git checkout master
% git checkout -b pb/new-branch
% git checkout -p as/version-logic
```

With `-p`, Git would show me each chunk of difference at a time and ask me if i should apply it, ignore it, or edit it then apply. I would do this once, keeping the things I want to be part of the first batch. For much chunks it's an easy yes-no, for some files touched by multiple distinct changes (e.g. `README` or `CHANGELOG`) it will require some editing, either as part of `-p` or after. Then commit, then repeat the `checkout -p` for the next batch, and so on.

If you'd like to try that, feel free. If not, just let me know when this branch has settled and I'll go ahead.

atreyasha:
Ah I understand. Let me try this out, will probably help me to try this out :)

Should I submit each commit as a separate PR/branch, or just 4 commits into one new PR?

pbrisbin:
> Should I submit each commit as a separate PR/branch, or just 4 commits into one new PR?

That's actually up to you. I would do 4 separate PRs, personally, because then things will get into master quicker, since we'll review and merge each increment as we go.

pbrisbin:
It can be annoying to do the PR-from-PR workflow though, so 4 commits in one PR is fine with me -- or even force-pushing the updated commits to this PR.

pbrisbin:
This has been replaced by #114, which was then itself replaced by #115, #116, and #117.

PR #114, opened 2020-05-04
Cleaned version of #108

atreyasha:
#108 reduced to 4 commits

### Remaining changes

1. Update the changelog based on how we want to bump to the next version. I just arbitrarily wrote the next version as `v6.4.0`.

2. Fine-tune `Makefile` as mentioned in #108.


pbrisbin:
It looks like things got a little jumbled, if you look at the first commit for example:

https://github.com/pbrisbin/downgrade/pull/114/commits/cabcba9f5e3443efc99d18de8b7c8f53c234cc67

It's including things in the CHANGELOG ffrom the last commit (version filtering), as well as manpage updates for the 3rd commit (removing `--arch`).

atreyasha:
Messed something up again. Will do it one more time.

atreyasha:
Alright, it looks good now from my end

Makefile
```diff
@@ -30,8 +30,12 @@ man: doc/downgrade.8
 test:
 	cram test
 
+.PHONY: config
+config:
```
pbrisbin, line 34: I think this should not be a `PHONY`, as it's building `downgrade` from `dowgrade.in`

```make
downgrade: downgrade.in
        sed ... $< > $@
```

And all the other tasks that you made depend on `downgrade.in`, they should depend on `downgrade` IMO, so that is is brought up to date and used in its interpolated form whenever needed.

Also, I usually use `m4` for this, but I honestly don't know why. It just felt more neck-beard-y.

test/helper.sh
```diff
@@ -24,4 +24,4 @@ write_pacman_conf() {
 ignore() { yes | prompt_to_ignore "$@" >/dev/null; }
 
 # Load implementation
-source "$SRCDIR"/downgrade
+source "$SRCDIR"/downgrade.in
```
pbrisbin, line 27: I don't know that this will work, or if it does I don't think it would always. I think this needs to continue to use `downgrade`, and the test just needs to depend on the `downgrade` target.

CHANGELOG.md
```diff
@@ -1,13 +1,17 @@
-## [*Ongoing*](https://github.com/pbrisbin/downgrade/compare/v6.3.0...master)
 
-- [NEW] Handling hard dependencies during downgrading
 
-## [*v6.3.0*](https://github.com/pbrisbin/downgrade/compare/v6.1.0...v6.3.0)
+## [v6.3.1](https://github.com/pbrisbin/downgrade/compare/v6.3.0...v6.3.1)
```
pbrisbin, line 19: I don't think we need this entry. This is all internal stuff that users wouldn't care about, no need for it in the CHANGELOG.

I also wonder if we should generally keep CHANGELOG out of PRs for now. It seems strange to couple the PRs to release cadence. We can (and should) do whatever PRs we want, then update the CHANGELOG for release at the end (which I'm happy to own myself).

Makefile
atreyasha: Yes I had tried this earlier, my arguments for making it a phony is the following edge-case:

Suppose one is doing local development and already created `downgrade` from a previous run. Then, when running `make install` with a non-default prefix (such as `/opt`), `downgrade` will not be made since `downgrade.in` was not modified. Now, `downgrade` will be installed with the "wrong" prefix. I just find it safer to always be able to remake `downgrade`, so any changes in the `make` prefix will always be reflected.

If we don't do this, we would always have to remember to delete a previously made version of `downgrade` if we want to experiment installing to another prefix.

test/helper.sh
atreyasha: It works with `downgrade.in`, since it only needs to source the variables/functions. If you want to keep it as `downgrade`, then you would probably need to add the `config` phony as a dependency to the `test` phony.

CHANGELOG.md
atreyasha: Yes I don't have a preference for this. I just wanted to keep the changelog up-to-date. I am fine with your suggestion of squash-merging the first two commits, and then merging the third one.

We can handle the fourth commit afterwards.

Makefile
atreyasha: IMO the actual `downgrade` file is only important for installation. Everything else can work off `downgrade.in`.

CHANGELOG.md
atreyasha: I won't change anything else for now, but feel free to edit it as you find best.

Makefile
pbrisbin: That's an interesting case. It just feels wrong to be using `downgrade.in` for things and having a phony target that always updates `downgrade`, whether it needs it or not. I'll give this some thought :thinking: 

I think local testing with alternative `PREFIX` is super unlikely, but there's also no real reason for worrying about making this sufficiently lazy.

pbrisbin: You mentioned something that I found compelling,

> IMO the [`PREFIX`] file is only important for installation

I think this is telling me to prefer the other option, where we create a file at installation time which contains the appropriate `TEXTDOMAINDIR` value, and then we source it from `downgrade`. Is there a reason you were against that?

I think I mentioned `/etc/profile.d/downgrade.sh` originally when I suggested that, but I'd also be fine with `$(PREFIX)/downgrade/textdomaindir.env` or something too.

atreyasha: Ah yes you mentioned this. My issue with that was that it involves another file being installed. But then again this would fix the whole conundrum with `downgrade.in`.

Sounds like a better alternative. We can add the command to create this within the `install` phony, so it is all compact and we don't need another rule. We can update the `uninstalll` phony to remove it too. Just to be safe, we could add a condition in (the new) `downgrade` where `TEXTDOMAINDIR` defaults to some predefined value in case `/etc/profile.d/downgrade.sh` cannot be found/sourced, just to ensure the script does not fail.

Alright, I will sign off for the day. Need to powerpoint an e-pub-quiz for getting drunk with friends tomorrow!

pbrisbin:
Haha nice. I can take care of that, just wanted to run it by you before
going ahead. Thanks!

pbrisbin:
:wave: Good morning. I've broken out 3 PRs and will now close this one:

1. #115 fixes `TEXTDOMAINDIR` in a much easier way, I'd love your thoughts (and testing) here
2. #116 is the removal of `--arch`; it's your commit, but I reduced the diff by not including a lot of locale file updates. once (1) is merged, I will rebase and deal with locale-file conflicts as well as run the new `make` target to fix the files overall
3. #117 is the version filtering; once (1) and (2) are merged, I can rebase this one and we can do more focused review on this one

I kept conflicting `CHANGELOG` updates in each PR, which just add whatever item(s) that PR is for to the `Ongoing` section. This will be annoying on merge, but I'm rolling with that workflow for now. I think in the future we should not touch CHANGELOGs in the PRs. If we want to track CHANGELOG items as the work is going on, let's put it in the PR _descriptions_. Then I can go back over them and build the CHANGELOG all together on mater before releasing.

PR #115, opened 2020-05-05
Respect PREFIX in locale files' installation

pbrisbin:
**NOTE**: there are number of minor but distinct commits here, which I've pulled
from #114. Please consider reviewing this PR by paging through the Commits tab
one-by-one.

The important commits here are:

- 539160f Find TEXTDOMAINDIR relative to script

  Previously, downgrade hard-coded `/usr/share/locale`. This meant
  installation to an alternative `PREFIX` would still use `/usr` for the
  locale files, which is surprising.

  This change means that wherever downgrade is installed, it will find locales
  relative to itself. This is behavior neutral on Arch, since the PKGBUILD[1]
  uses an explicit `PREFIX` of `/usr`.

  A future commit will bring the locale files installation into the make install
  target, which will respect PREFIX. That will only function with the
  pre-requisite commit.

  [1]: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=downgrade

- 53b277b Handle locale files in make targets

  This was previously handled directly in the `PKGBUILD`, but there's no reason
  it can't be part of the general project targets. This ensures `PREFIX` is
  handled uniformly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
-PREFIX    ?= /usr/local
-MANPREFIX ?= $(PREFIX)/share/man
-PANDOC    ?= $(shell which pandoc)
+PREFIX       ?= /usr/local
+LOCALEPREFIX ?= $(PREFIX)/share/locale
+MANPREFIX    ?= $(PREFIX)/share/man
+PANDOC       ?= $(shell which pandoc)
 
 setup:
 	command -v cram   || aurget cram
@@ -27,12 +28,18 @@ install:
 	install -Dm644 doc/downgrade.8 $(DESTDIR)$(MANPREFIX)/man8/downgrade.8
 	install -Dm644 completion/bash $(DESTDIR)$(PREFIX)/share/bash-completion/completions/downgrade
 	install -Dm644 completion/zsh $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_downgrade
+	for po_file in locale/*.po; do \
+	  locale="$$(basename "$$po_file" .po)"; \
+	  mkdir -p "$(DESTDIR)$(LOCALEPREFIX)/$$locale/LC_MESSAGES/"; \
+	  msgfmt "$$po_file" -o "$(DESTDIR)$(LOCALEPREFIX)/$$locale/LC_MESSAGES/downgrade.mo"; \
+	done
 
 uninstall:
 	$(RM) $(DESTDIR)$(PREFIX)/bin/downgrade \
 	  $(DESTDIR)$(MANPREFIX)/man8/downgrade.8 \
 	  $(DESTDIR)$(PREFIX)/share/bash-completion/completions/downgrade \
-	  $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_downgrade
+	  $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_downgrade \
+	  $(DESTDIR)$(LOCALEPREFIX)/*/LC_MESSAGES/downgrade.mo
 
 .PHONY: setup test install uninstall
 
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,8 @@ install:
 uninstall:
 	$(RM) $(DESTDIR)$(PREFIX)/bin/downgrade \
 	  $(DESTDIR)$(MANPREFIX)/man8/downgrade.8 \
-	  $(DESTDIR)$(PREFIX)/share/bash-completion/completions/downgrade
+	  $(DESTDIR)$(PREFIX)/share/bash-completion/completions/downgrade \
+	  $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_downgrade
 
 .PHONY: setup test install uninstall
 
--- a/Makefile
+++ b/Makefile
@@ -23,16 +23,15 @@ test:
 	cram test
 
 install:
-	install -Dm755 downgrade $(DESTDIR)/$(PREFIX)/bin/downgrade
-	install -Dm644 doc/downgrade.8 $(DESTDIR)/$(MANPREFIX)/man8/downgrade.8
-	install -Dm644 completion/bash $(DESTDIR)/$(PREFIX)/share/bash-completion/completions/downgrade
-	install -Dm644 completion/zsh $(DESTDIR)/$(PREFIX)/share/zsh/site-functions/_downgrade
+	install -Dm755 downgrade $(DESTDIR)$(PREFIX)/bin/downgrade
+	install -Dm644 doc/downgrade.8 $(DESTDIR)$(MANPREFIX)/man8/downgrade.8
+	install -Dm644 completion/bash $(DESTDIR)$(PREFIX)/share/bash-completion/completions/downgrade
+	install -Dm644 completion/zsh $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_downgrade
 
 uninstall:
-	$(RM) $(DESTDIR)/$(PREFIX)/bin/downgrade \
-	  $(DESTDIR)/$(MANPREFIX)/man8/downgrade.8 \
-	  $(DESTDIR)/$(PREFIX)/share/bash-completion/completions/downgrade \
-	  $(DESTDIR)/$(PREFIX)/share/zsh/site-functions/_downgrade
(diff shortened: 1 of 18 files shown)


Makefile
```diff
@@ -13,28 +15,37 @@ locale/downgrade.pot: downgrade
...
-	  $(DESTDIR)/$(MANPREFIX)/man8/downgrade.8 \
-	  $(DESTDIR)/$(PREFIX)/share/bash-completion/completions/downgrade \
-	  $(DESTDIR)/$(PREFIX)/share/zsh/site-functions/_downgrade
-
-.PHONY: setup test install uninstall
+	$(RM) $(DESTDIR)$(PREFIX)/bin/downgrade \
+	  $(DESTDIR)$(MANPREFIX)/man8/downgrade.8 \
+	  $(DESTDIR)$(PREFIX)/share/bash-completion/completions/downgrade \
+	  $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_downgrade \
+	  $(DESTDIR)$(LOCALEPREFIX)/*/LC_MESSAGES/downgrade.mo
```
atreyasha, line 48: This is a nice touch to combine all of the installation with locales in one place. Also my bad, I missed the uninstall issue for zsh completions.

downgrade
```diff
@@ -384,5 +382,9 @@ cli() {
 
 if ((!LIB)); then
   set -e
+
+  TEXTDOMAINDIR=$(cd "$(basename "$0")"/../share/locale && pwd)
+  export TEXTDOMAIN=downgrade TEXTDOMAINDIR
```
atreyasha, line 387: Also a nice touch. This approach removes the need for any `in` file and any extra configuration files as well. It is also a fair assumption for the locale directory to be located adjacent to where `downgrade` is installed, especially given that is the approach in our package-build.

Makefile
atreyasha: > A future commit will bring the locale files installation into the make install
target, which will respect PREFIX. That will only function with the
pre-requisite commit.

This sounds like a good addition. Maybe we could also bring man-doc building into that. But we can discuss that in a future PR.

downgrade
atreyasha: When running `downgrade vim`, I get the following error:

```
/usr/local/bin/downgrade: line 386: cd: downgrade/../share/locale: Not a directory
```

I guess this is because `basename "$0"` would evaluate to `downgrade`. We could change `basename "$0"` to `which "$0"`, but this would evaluate to `/usr/local/bin/downgrade` and it will not be possible to `cd` on this as this is a file and not a directory.

I propose the following change which I also tested locally to replace line 386:

```shell
TEXTDOMAINDIR=$(realpath $(dirname $(which "$0"))/../share/locale)
```

Makefile
```diff
@@ -23,16 +23,15 @@ test:
...
+	install -Dm644 completion/zsh $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_downgrade
 
 uninstall:
-	$(RM) $(DESTDIR)/$(PREFIX)/bin/downgrade \
-	  $(DESTDIR)/$(MANPREFIX)/man8/downgrade.8 \
-	  $(DESTDIR)/$(PREFIX)/share/bash-completion/completions/downgrade \
-	  $(DESTDIR)/$(PREFIX)/share/zsh/site-functions/_downgrade
+	$(RM) $(DESTDIR)$(PREFIX)/bin/downgrade \
+	  $(DESTDIR)$(MANPREFIX)/man8/downgrade.8 \
+	  $(DESTDIR)$(PREFIX)/share/bash-completion/completions/downgrade
```
atreyasha: I agree with the idea here to make it cleaner. But the `PREFIX` defined in this `Makefile` has no trailing slash symbol so there is an inconsistency between `DESTDIR` and `PREFIX`. If we want to keep this, `PREFIX` should be `/usr/local/` and line 26 (as an example) would become `install -Dm755 downgrade $(DESTDIR)$(PREFIX)bin/downgrade`.

I would be uncomfortable with this version because if a trailing slash is erroneously forgotten during say manual installation, it would lead to a wrong prefix and wrong (perhaps non-existant) installation path. However, if we leave the structure as before and allow for repeating slashes (even when unnecessary), we can be sure that this error would never arise and `downgrade` will always be installed to the correct (intended) location.

downgrade
pbrisbin: That was silly, it should just be `dirname`, not `basename`. using `realpath` vs `&& pwd` doesn't matter, so I'm inclined to keep my style.

Makefile
atreyasha: > Path variables should have leading slashes themselves (like the default
PREFIX), so $(DESTDIR)/$(PREFIX) was actually "//usr/local". This
doesn't matter, but is unnecessary.

Ah just read this commit message. Hmm, yes if we are sure the `DESTDIR` will always have a leading slash this should be fine. But if want to play it safe we could keep the slash symbol between `DESTDIR` and `PREFIX`. Let me know your thoughts :)

downgrade
atreyasha: Yes, tested just replacing `basename` with `dirname` and it works. Marking this as resolved.

atreyasha: Sorry to open this conversation again, I came across another error. Changing `basename` to `dirname` works find for the installed file. However, this creates a problem for us during development. So let's say I made some changes to `downgrade` and want to test it from my local repo. Say I run `./downgrade vim`. Then I get the following error message:

```
./downgrade: line 386: cd: ./../share/locale: No such file or directory
```

Perhaps we add some error handling for `TEXTDOMAINDIR` to default to `/usr/share/local` if the output of the `cd` command is non-zero?

atreyasha: Or alternatively:

```shell
if [ -f "$(dirname "$0")"/../share/locale ]; then 
    TEXTDOMAINDIR=$(cd "$(dirname "$0")"/../share/locale && pwd)
    export TEXTDOMAIN=downgrade TEXTDOMAINDIR
fi
```

Makefile
pbrisbin: Nice catch! I'm not sure what is least surprising to a packager. I think it's worth enumerating the cases (and possibly writing a small cram test for it?):

What do you think of these outcomes, and can you list any more interesting combinations?

```
make DESTDIR= PREFIX= install
# should install bin/downgrade or /bin/downgrade?

make DESTDIR=/d PREFIX= install
# should install /d/bin/downgrade

make DESTDIR=d PREFIX= install
# should install d/bin/downgrade

make DESTDIR= PREFIX=/p install
# should install /p/bin/downgrade

make DESTDIR= PREFIX=p install
# should install p/bin/downgrade

make DESTDIR=/d PREFIX=/p install
# should install /d/p/bin/downgrade

make DESTDIR=d PREFIX=/p install
# should install d/p/bin/downgrade

make DESTDIR=d PREFIX=p install
# will install dp/bin/downgrade
# sorrynotsorry
```

> I would be uncomfortable with this version because if a trailing slash is erroneously forgotten...

FWIW, since this is a _packager_ interface and not a user interface (yes, someone installing from source plays both roles at different times, but they're still different roles), I don't mind there being risks if you get the slashes wrong. The complexity of making it totally free of foot-guns isn't there IMO. So if we come up with a nice simple approach that solves the cases above, but maybe could be misused accidentally somehow, that'd be fine with me.

downgrade
pbrisbin: Ah yes, we should do something for in-repo `./downgrade` usage. I ran into this error with the test and fixed it by putting this under the `LIB` check. Can you tell I almost exclusively develop by tests?

Makefile
pbrisbin: > Let me know your thoughts :)

I'm going to keep this as-is for now, but I'm not against revisiting in a separate PR and improving things.

Makefile
pbrisbin: > Maybe we could also bring man-doc building into that. But we can discuss that in a future PR.

Agreed. My only concern would be bringing in a `pandoc` makedepend, it's a bit heavy.

downgrade
pbrisbin: I think this is fixed now.

Makefile
atreyasha: Good morning :sun_behind_small_cloud:

Did not have enough time to think about this yesterday. I agree that this is minor and we can discuss this in a future PR if need be.

PR #116, opened 2020-05-05
Remove --arch

pbrisbin:
Extracted from, and replacement to, #114.

PR #117, opened 2020-05-05
Implement version filtering

pbrisbin:
Extracted from, and replacement for, #114.


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## [*Ongoing*](https://github.com/pbrisbin/downgrade/compare/v6.3.0...master)
 
+- [NEW] Version bound handling by parsing logical =,>,<,>= and <= operators
+- [NEW] Handling basic version conflicts during downgrading
 - [NEW] Handling hard dependencies during downgrading
 
 ## [*v6.3.0*](https://github.com/pbrisbin/downgrade/compare/v6.1.0...v6.3.0)
```
atreyasha, line 7: Same minor comment as #116: there might be a merge conflict here given that version headers are bolded, but in the previous PR they are not.

pbrisbin:
@atreyasha I just pushed a simpler implementation that is less disruptive to the surrounding code (and so less risky). 

At the end of the day, all we're doing now is:

```diff
- candidates=$(search | sort)
+ candidates=$(search | filter | sort)
```

Please give this a test and let me know if it works.

pbrisbin:
Ah good catch.

I expected there to be more `vercmp` fixtures, because I changed `filter_` to always presumptively call it, instead of only if there was a comparison operator in the arguments. However, it should certainly not continue to generate new fixtures every run. This should be easy to track down and resolve by running the test with `FIXTURES_NO_RECORD=1`. With this flag, the test suite would error instead of recording any new fixtures. However, I swore we used this flag on CI, so I'm not sure how things are Green there.

I don't have a moment now, but that's where I'd start investigating.

test/main/one-choice.t
```diff
@@ -5,7 +5,8 @@ It doesn't present packages when there's only one choice
 
   $ touch "$TMP/foo.pkg"
   > search_packages() { printf "%s\n" "$TMP/foo.pkg"; }
```
atreyasha, line 7: The `vercmp` fixture issue arose from this line in `test/main/one-choice.t` and from line 7 in `test/main/success.t`.

Because of `search_packages() { printf "%s\n" "$TMP/foo.pkg"; }`, a randomly created `$TMP` string from `cram` would be fed in to `main`, which would ultimately trickle down as an argument of the `act-like` version of `vercmp`. Since the arguments to this mocked `vercmp` will always have randomly generated prefix, it would always search for a different `md5sum` and would never find it, and that's why it would keep creating it.

I added a commit to fix this and it seems to work fine now. Would be great if you could test it on your system as well.

pbrisbin: I see, that mostly makes sense and the fix looks good. I'm a little concerned we're still missing something though.

The real `filter_packages` function has a condition for empty `operator` and `version` and should just `return 0` immediately (without calling `vercmp`) if they are empty. These tests call `main foo`, so why is this condition not being met?

Also, the one test you changed (`one-choice.t`) is actually dependent on how `filter_packages` behaves, so I'm worried that mocking it out like that is risky.

pbrisbin: And a minor note for next time: it'd be good to always commit any fixture changes in their own commit. GitHub was struggling to render that diff :joy:

atreyasha: > The real filter_packages function has a condition for empty operator and version and should just return 0 immediately (without calling vercmp) if they are empty. These tests call main foo, so why is this condition not being met?

That's a really good point. Just looked and saw that the conditional check for no operator or version is inside `matches_name_version_filter`, which means the `vercmp` is already executed one line before it. Maybe we move the conditional check for no operator or version one level up to `filter_packages`?

atreyasha: > And a minor note for next time: it'd be good to always commit any fixture changes in their own commit. GitHub was struggling to render that diff

Aha yes sorry about that.

pbrisbin: I reverted your fix and applied mine, WDYT?

atreyasha: Oh shit, I just applied a fix and did not see this message, so maybe it overrode it. Maybe force push it?

atreyasha: And another thing, I noticed that the phony `locale` is no longer present in the `Makefile`. Could we add the following line to `Makefile`:

```
.PHONY: locale
locale: locale/downgrade.pot
```

This just makes it easier to locally update the locale, and is also consistent with how we did it for `make man`. I will not change anything in case I conflict with your commits.

pbrisbin:
No worries. History should be clean now.

pbrisbin:
Released as v7.0.0 :tada:  nice job!

pbrisbin:
I forgot CHANGELOG, but will update eventually.

atreyasha:
> Released as v7.0.0 tada nice job!

Awesome, enjoying this release cycle :smile:

Wishing you a nice day ahead!

Issue #118, opened 2020-05-07
Non-interactive downgrading with version-filtering


atreyasha:
Hi @pbrisbin,

Making a new issue for a new idea. Besides the current downgrading dependencies idea in #83, I would like to implement an additional command line option for non-interactive functionality, and this option can be called `--non-interactive`.

Basically, this option will enable lazy users to downgrade package(s) with just some basic instructions and without having to select any option from the table. Since `downgrade` is by design an interactive tool, this option would be non-default. Some examples of this option's functionality using our staple `vim`:

1. `downgrade vim=8.2 --non-interactive`: this will result in the latest version after fuzzy match to be automatically selected and installed.

2. `downgrade "vim(<=|>=|<|>)8.2" --non-interactive`: this will result in either the latest version before `(<|<=)` or the latest version after `(>|>=)` to be automatically selected and installed.

**Note:** For all of these options, first priority is install via cache, if not via ALA.

I find this option useful because it enables `downgrade` to be used without any user input, so users who want to write scripts to return their system to a certain state could easily use this non-interactive tool. WDYT?

pbrisbin:
I definitely like this idea!

Generally, I would like to push on one of our existing core behaviors, which is:

> If you can invoke `downgrade` such that there's only one option, it's handled non-interactively

What do you think of supporting a `~` at the end of any operator,

```sh
downgrade vim=~8.2
# filter to the latest version in 8.2.x

downgrade vim>~8.2
# filter to earliest version after 8.2 (exclusive)

downgrade vim>=~8.2
# filter to earliest version after 8.2 (inclusive)

downgrade vim<~8.2
# filter to latest version latest version before 8.2 (exclusive)

downgrade vim<=~8.2
# filter to latest version latest version before 8.2 (inclusive)
```

And, as always, if this filtering reduces to a single option, it is handled non-interactively.

> For all of these options, first priority is install via cache, if not via ALA.

Again, I'd like to establish and rely on a core principal, instead of defining new and specific behavior. What if, the following were just how it always worked:

> If ever the same package-version appears in ALA and Cache, only the Cache option is presented

That seems like a reasonable and useful truism of `downgrade` to me. And if it were true, the situation you describe just always applies, rather than being something special to a new `--non-interactive` option. And it ensures that if you filter down to the same option in Cache and ALA, only the Cache option is shown, and oh hey that's just one option, so it's handled non-interactively.

[...]

PR #121, opened 2020-05-09
Minor doc-fix, proposed modification to "=" operator and addition of "=~" wrapper operator

atreyasha:
### 1. Minor man-doc fix

* c87b140: modification to man-doc to sort the hierarchies of man-pages

### 2. Minor fix *assuming we keep the current `=` operator*

* c87b140...0a306a6: minor bugfix for current `=` operator in case regex does not match and `vercmp` does match. This could be the case where `8.2.33` would not regex match `8.2.0033`, but it would still be a `vercmp` match.

* 0a306a6...085b4b6: when inputting a (failing) version bound such as `downgrade vim=1.2`, the error message will state `Unable to downgrade vim=1.2`. I added a slight modification such that the error message would say `Unable to downgrade vim` to show that `downgrade` was able to parse the actual package.

### 3. Major fix *assuming we modify the current `=` operator*

>I found it surprising (that = doesn't just mean =), but I suppose the unexpected behavior would result in multiple results and present the list to the user, so it's not particularly problematic.

I was thinking about your comment in #118 and came to realize that the `=` operator should indeed really mean equals and nothing else. To address this, I reverted the original `=` operator to have the correct meaning and added a new `=~` operator which behaves exactly like that in bash (it just a wrapper for that). This will provide the user with the ability to conduct regex on package versions directly from the command line, for example `downgrade "vim=~\.0033"`.

To address this, I chunked similar commits as per (2):

* 085b4b6...0cbd8a3: revert `=` operator to equal meaning and add regex matching operator `=~`. Add respective tests for this as well.

* 0cbd8a3...99f35d4: when inputting a (failing) version bound such as `downgrade vim=~1.2`, the error message will state `Unable to downgrade vim=~1.2`. I added a slight modification such that the error message would say `Unable to downgrade vim` to show that `downgrade` was able to parse the actual package.

### Commits chunking

In this PR, (2) and (3) are mutually exclusive so we should choose what we like best. I would personally go for (3) since it addresses your comment in #118 and also provides the user with more flexibility in filtering packages while keeping the meaning of `=` to a hard equal.

If we like (2) more than (3), we can just drop the commits from 4868622 onward. If we like (3), we can just merge everything since the commits in (2) will be directly overriden.

I hope the chunking of commits helps in selecting what we find best. I tried to structure it as independently as possible so we can just pick and choose.


downgrade
```diff
@@ -305,6 +305,7 @@ declare -a to_ignore
 declare -a to_install
 declare -a pacman_args
 declare -a installed
+declare current
```
pbrisbin, line 312: Nice fix. I have a sneaking suspicion that `installed` and `current` could be made local to the `present_packages` function, but it would have to be re-imagined a little bit with something clever like building a `sed` that replaces current and installed package names with their decorated versions, so you could feed `output_package` through it. :thinking:

downgrade
```diff
@@ -271,6 +271,7 @@ main() {
   for term; do
     if ! process_term "$term"; then
       {
+        term=$(sed -r "s/(.*)\b(>=|<=|<|>|=)\b(.*)/\1/g" <<<"$term")
```
pbrisbin: Right now, this operator-version "parsing" happens exactly once, within `process_term`. If we start doing it in more places, we run the risk of it going out of sync. For such a minor, not-happy-path feature like this error message, I wouldn't to it, personally.

If you want to keep it, that's fine. I might also consider the following two options:

1. Print this error message from within `process_term`, before it returns `1`, where `$name` is already available
2. Extract this `sed` to a function and re-use it

downgrade
atreyasha: Thank you. Hmm, yes this could be a fix to reduce the number of global variables. We could address this in another PR.

downgrade
atreyasha: Somehow I am still clinging on to this error message, because I feel like if it gives you an output with one of the operator symbols, it should already be telling you that it was unable to parse the operator. If it is fine with you, I will keep this and follow your suggestions.

atreyasha:
> LGTM.
> 
> In the future, let's not update the rendered man page or update the locales files in PRs; it's just diff noise that slows down the GitHub UI :). I will do all that as part of release, along with the CHANGELOG, etc.

Got it. I will edit the commits now just to ensure there are no conflicts for the other PR. Just to check, so for this PR you would prefer option (3) over option (2)? In that case, I can simplify the commits too.

pbrisbin:
> for this PR you would prefer option (3) over option (2)?

Oh sorry, I forgot there were choices and just reviewed the PR as-is (which I guess is (3)?). I don't have a strong opinion, I'd defer to you. If pressed, I guess (3).

downgrade
atreyasha: Took option 1 in this case to keep the `sed` operator parsing to just one place so we don't need to add another function now.

atreyasha:
Reduced everything to 4 commits. Changes in locale and man-doc are now ommitted to prevent. If everything is fine, you can merge it.

Issue #120, opened 2020-05-08
Creating --pacman-log, --pacman-cache and --maxdepth command-line arguments


atreyasha:
As per #119, there are minor issues which we can address.

1. Create `--pacman-log` and `--pacman-cache` command line options which can take space-separated paths to identify single or multiple files. This is consistent with #96 in order to represent environmental variable as command-line options.

2. Investigate whether `maxdepth` is necessary in `downgrade` for finding local cached files. Or possibly augmenting `maxdepth` as a command-line option such as `--maxdepth`. This could be useful in non-default cache locations.

PR #123, opened 2020-05-10
Add --pacman-cache, --pacman-log and --maxdepth to CLI options

atreyasha:
This PR addresses #120 by adding `--pacman-cache`, `--pacman-log` and `--maxdepth` as CLI options. Commits have been chunked to represent discrete components of this PR.

Regarding whether to include `--maxdepth` at all, I think it is a good idea to include this parameter to ensure that `find` does not get stuck searching some deep user-defined cache directory. The user can add an arbitrarily large value for this if they want to search some basic cache directory for (eg.) AUR packages.

### DIFF
--- a/completion/zsh
+++ b/completion/zsh
@@ -2,9 +2,12 @@
 
 _downgrade(){
   _arguments -S -C '*:packages:->pkg' \
-             '--pacman[Specify pacman command]' \
-             '--pacman-conf[Specify pacman configuration file]:config:_files' \
-             '--ala-url[Specify location of ALA server]' \
+             '--pacman[Specify pacman command]:pacman command' \
+             '--pacman-conf[Specify pacman configuration file]:pacman config file:_files' \
+             '--pacman-cache[Specify pacman cache directory or directories]:pacman cache directory:_path_files -/' \
+             '--pacman-log[Specify pacman log-file]:pacman log file:_files' \
+             '--maxdepth[Specify maximum depth for local cache find]:maximum search depth' \
+             '--ala-url[Specify location of ALA server]:ala url' \
              '--ala-only[Option enables only ALA server search]' \
              '--cached-only[Option enables only cached search]' \
              '--nosudo[Option enables only su even when sudo is present]' \
--- a/doc/downgrade.8.md
+++ b/doc/downgrade.8.md
@@ -60,15 +60,25 @@ The columns have the following meaning:
 
 # OPTIONS
 
-## DOWNGRADE OPTIONS
-
 **\--pacman** *\<command\>*\
 
 > Pacman command, default is *pacman*.
 
-**\--pacman-conf** *\<file-path\>*\
+**\--pacman-conf** *\<path\>*\
   
 > Pacman configuration file, default is */etc/pacman.conf*.
+
+**\--pacman-cache** *\<path\>*\
+
+> Pacman cache directory or directories, default value is extracted from pacman configuration file, or otherwise defaults to */var/cache/pacman/pkg*. Multiple cache directories can be supplied as space-separated paths.
+
+**\--pacman-log** *\<path\>*\
+
+> Pacman log file, default value is extracted from pacman configuration file, or otherwise defaults to */var/log/pacman.log*.
+
+**\--maxdepth** *\<int\>*\
+
+> Maximum depth to search for cached packages, defaults to *1*.
 
 **\--ala-url** *\<url\>*\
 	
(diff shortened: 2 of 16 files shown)


downgrade
```diff
@@ -159,7 +169,7 @@ search_packages() {
     : "${PACMAN_CACHE:=/var/cache/pacman/pkg/}"
 
     # shellcheck disable=SC2086
-    find $PACMAN_CACHE -maxdepth 1 -regextype posix-extended -regex ".*/$pkgfile_re"
+    find $PACMAN_CACHE -maxdepth $MAXDEPTH -regextype posix-extended -regex ".*/$pkgfile_re"
```
pbrisbin, line 172: `shellcheck disable=SC2086` is here because we _want_ `$PACMAN_CACHE` to word-split if necessary, but `$MAXDEPTH` can (and should) be quoted.

downgrade
```diff
@@ -299,6 +309,7 @@ main() {
 : "${DOWNGRADE_FROM_ALA:=1}"
 : "${DOWNGRADE_FROM_CACHE:=1}"
 : "${DOWNGRADE_NOSUDO:=0}"
+: "${MAXDEPTH:=1}"
```
pbrisbin, line 312: Please use `DOWNGRADE_MAXDEPTH`, to be consistent with the others. I don't like tools that expect un-namespaced environment variables, as there can be collisions or just confusion.

doc/downgrade.8.md
```diff
@@ -92,18 +106,14 @@ As per the usage syntax, any options supplied after the **\--** character sequen
 
 ## DEFAULT BEHAVIORS
 
-By default, **downgrade** will search both local caches and the ALA.
+By default, **downgrade** will search both local caches and the ALA.
```
pbrisbin, line 109: :art: you introduced some trailing whitespace here.

doc/downgrade.8.md
```diff
@@ -60,20 +60,34 @@ The columns have the following meaning:
 
 # OPTIONS
 
-## DOWNGRADE OPTIONS
+## DOWNGRADE OPTIONS WITH ARGUMENTS
```
pbrisbin, line 63: FWIW, I don't it particularly valuable to have distinct sections.

completion/zsh
```diff
@@ -4,6 +4,9 @@ _downgrade(){
   _arguments -S -C '*:packages:->pkg' \
              '--pacman[Specify pacman command]' \
              '--pacman-conf[Specify pacman configuration file]:config:_files' \
+             '--pacman-cache[Specify pacman cache location(s)]:config:_files' \
+             '--pacman-log[Specify pacman log-file location]:config:_files' \
+             '--maxdepth[Specify maximum depth for local cache find]' \
```
pbrisbin, line 9: Do you have to do something here to indicate it takes an argument?

doc/downgrade.8.md
atreyasha: Woops, fixed.

doc/downgrade.8.md
atreyasha: Will revert back.

downgrade
atreyasha: I see, will do.

downgrade
atreyasha: Done.

completion/zsh
atreyasha: As far as I know, there is no autocomplete prompt for integers here. We can prompt it to complete files or paths as per the options above, but that is not the case for `--maxdepth`.

atreyasha:
Made the respective fixes. Removed the commits for the man-doc to prevent conflicts. I had to keep the changes to the locale since I added translations for new keys. If all is fine, you can merge it.

completion/zsh
pbrisbin: Well, I wasn't expecting it to be smart about completing integers, I was just afraid that if you did

```
% downgrade --maxdepth <tab>
```

and it doesn't know that `--maxdepth` take _some_ argument, it'll start completing more `downgrade` options at that position, instead of stopping like actual `find` does:

atreyasha: Ah I understand now! That should not be hard to do. Let me look through something and update the PR.

atreyasha: Fixed, now it will not complete any option for `--maxdepth` and `--pacman`. I modified such that it will complete directories and not files for `--pacman-cache`.

atreyasha:
Is it okay if I do the merging to master and fixing of the conflicts directly in master?

pbrisbin:
> I do the merging to master (for both PRs) and fixing of the conflicts directly in master?

You should be able to,

- Merge one of them (GitHub UI shows no conflicts)
- Rebase the other one, fix conflicts
- Merge the other one

I don't think anything should need fixing "in master", right?

atreyasha:
Ah sorry I worded that wrongly, there should be no conflicts in `master` itself. There will probably be some conflicts before the merge but I can fix those in the UI directly. Will do it then.

atreyasha:
Merging done for both. Just need to rebuild the man-page/changelog which I will leave to you.

pbrisbin:
And just double-checking, major version bump because we changed how `=` works?

atreyasha:
Yes major version bump sounds right.

pbrisbin:
:tada: 8.0.0 released

PR #127, opened 2020-07-27
Clarify the default action for IgnorePkg prompt

132ikl:


atreyasha:
Thank you for the PR. From my POV, this looks good. If we merge this, we would need to run `make locale` and then manually update the `./local/*.po` files with the same uppercasing for the negative option in the other languages.

We can get feedback from @pbrisbin and then figure out the next steps. If it is fine with everyone, I would also be happy to update the locales.

pbrisbin:
Makes sense to me. @atreyasha feel free to merge and handle any follow-up tasks. Just let me know when to release a new version to the AUR.

PR #128, opened 2020-07-29
Ignore prompt default option clarification

atreyasha:
Pull request to merge changes from #127 along with locale updates.

### DIFF
--- a/downgrade
+++ b/downgrade
@@ -90,7 +90,7 @@ prompt_to_ignore() {
   for pkg; do
     grep -Eq '^IgnorePkg.*( |=)'"$pkg"'( |$)' "$PACMAN_CONF" && return 0
 
-    eval_gettext "add \$pkg to IgnorePkg? [y/n] "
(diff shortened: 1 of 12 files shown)


atreyasha:
@pbrisbin: merged this change. I would say we could now do a minor version bump because of several changes since the last release (https://github.com/pbrisbin/downgrade/compare/v8.0.0...master).

pbrisbin:
I don't see any new functionality there, it seems mostly internal stuff and minor copy changes, so I'm inclined to go with only a patch bump, 8.0.1 -- WDYT?

atreyasha:
Yes, while most of the diffs are indeed patch-worthy, I would argue commit e30e90c involves a behaviour change since the additional `-L` flag prepended to `find` would now allow for `find` to search through symlinks; whereas this was not possible earlier. This is why I think am inclined to a minor version bump. WDYT?

pbrisbin:
Eh, I can't decide if that is a Bugfix or a Feature. But I don't feel strongly, so 8.1.0 it is.

PR #130, opened 2020-08-08
Add completions for fish shell

Jakeler:


atreyasha:
Thank you for the PR. LGTM overall. I don't use `fish` but just did a minimal install and tested, and the completions work well. Let's get feedback from @pbrisbin and we can take it from there.

atreyasha:
Merged. Patch version bump?

pbrisbin:
I guess it depends if you call it a new feature, and so minor. I'm about to go on vacation, so we've got a little time to decide. Also, we should set you up to handle releases soon :)

atreyasha:
Nice :) Hope it is as warm on the other side of the world hehe

> Also, we should set you up to handle releases soon :)

That would be awesome. I believe it is possible to add co-maintainers for AUR packages (directly through the website on the right sidebar on the package's webpage). My AUR username is `atreyasha` or [here](https://aur.archlinux.org/account/atreyasha)

> I guess it depends if you call it a new feature, and so minor.

Hmm, IMO I don't see it as a new feature because `downgrade` itself has not changed and to whichever extent things have changed; only `fish` users would benefit from this. So I would still see it as a patch. WDYT?

Issue #131, opened 2020-08-12
Makefile patches


pbrisbin:
Following up from https://github.com/pbrisbin/downgrade/pull/130#issuecomment-670925268, I'd like to see @atreyasha handle the next release.

I added you as co-maintainer in the AUR. For release, I typically do the following:

1. Ensure translation files are all set
2. Ensure `CHANGELOG.md` is correct
3. Run `make release.patch` (in this case)

The `release` task relies on two other pieces of tooling:

- [`vbump`](https://aur.archlinux.org/packages/vbump-git/): a *very simple* package that just bumps a selected component of version numbers
- [`aur-release`](https://github.com/pbrisbin/dotfiles/blob/master/local/bin/aur-release): a shell script for pushing to the AUR repository, handling `SRCINFO`, sha sums, and all the rest

I apologize it's not quite as polished and there are a few desperate parts to this. Please do feel free to use (or create) your own process, I just wanted to make you aware of the tooling I had in case it's useful.

Any questions, let me know!

atreyasha:
Awesome, thank you. I got the notification for the AUR as well. Your tools are very helpful and I sourced/adapted them before for another AUR project.

Quick question: Do I have permissions to commit directly to `master` for the release? Last I checked, it was a protected branch and I was not allowed to commit directly.

pbrisbin:
Ah good point! You're right, you can't currently push directly to `mater`.

Let me know if you have any others, but I think our options are:

1. Make you an Admin on the repository, so you can commit around branch protection
2. Remove branch protection
3. Make all changes (including version bump and CHANGELOG) via PRs

I don't want to do (2) if we can avoid it. After a few releases, I think we could do (1), but maybe not yet. So how about (3) for now?

This means I'd have to Approve your PR that finalizes the release (not necessarily a bad thing for now), then you would merge the PR, create the tag, and proceed with the AUR repository update. I *think* you'll be able to create tags, even with the branch protection in place and your current non-admin status.

atreyasha:
Makes sense to do (3) for the next few releases so I can get the hang of things, then we could do (1) in the future. Will open the (release) PR soon.

atreyasha:
So it all worked. For future releases, we can follow the same procedure:

1. Open a PR for the changelog/locales
2. Merge to `master`
3. Run `make release.*`, which will push tags directly to `master` and push a new commit to the AUR.

Just some comments based on my fiddling around with `Makefile` in the last 30 mins:

[...]

PR #132, opened 2020-08-20
Release v8.1.1

atreyasha:
PR for patch version bump to `v8.1.1`, address #131. After the merge, will create the tag via GitHub UI (should be possible) and then update the AUR.


atreyasha:
It works as well via CLI, so will just use that.

PR #139, opened 2020-09-21
Add ISSUE_TEMPLATE

### LINKED ISSUE #137: GitHub issue templates
Currently, we do not use an issue template. This issue proposes the introduction of an issue template which lists certain tasks that should be completed before filing an issue. These could include:

1. Ticking a box the user checked for similar issues and could not find any
2. Listing `downgrade` version
3. Running `bash -x` on the `downgrade` command that was running into errors, if any

**PS**: Would it be useful for us to have a `downgrade --version` CLI option, such that users could easily check the version without having to go through `pacman`? I personally like the idea and think this could be another issue worth opening.

pbrisbin:
Closes #137.


atreyasha:
I just realized that the issue templates actually live in the repository, I had thought before that they were private in the UI. In any case, I was able to edit them further and separate them into three templates; specifically for bug reports, feature requests and questions/help. I also confirmed that users (or perhaps just us) can still use a blank template, for example when we are creating development-related issues.

I will approve this and create a new request for yourself.

atreyasha (approved):

pbrisbin:
Awesome thanks!

Issue #135, opened 2020-08-30
Multiple pacman cache directories in the form of aggregated CLI options


atreyasha:
Currently, the only way of passing multiple pacman cache directories would be to concatenate them into a space-separated string, such as:

```shell
downgrade --pacman-cache "/var/cache/pacman /my/thing/somewhere" foo
```

This issue proposes a modified feature where pacman caches can be aggregated, such that a user can specify multiple locations with ease:

```shell
downgrade
  --pacman-cache /var/cache/pacman \
  --pacman-cache ~/my/thing/somewhere \
  foo
```

PR #141, opened 2020-10-20
Pacman cache accumulation

atreyasha:
### Checklist

- [x] Update cram unit tests
- [x] Update mandoc

~- [ ] **Optional:** This might be an extreme edge case. Assume if the same package is present in two different pacman cache locations (maybe because it was an AUR package before and got shifted to one of the official repos later on). Based on local tests, I noticed that `downgrade` now would display the package twice in the menu. This could be solved by showing the location of the cache corresponding to the package in the menu, but I can imagine the menu would look horrible. Another option would be to just show one (unique) package if multiple copies are present. WDYT?~

### Description

This PR addresses #135 by modfying the `PACMAN_CACHE` variable read from stdin into a bash array. This allows us to search multiple cache locations with the `search_packages` function. Based on local tests, it also does not conflict with the case where there is only one pacman configuration location. Try out the following:

```shell
$ downgrade --pacman-cache /path/to/somewhere --pacman-cache /path/to/somewhere/else foo bar
```

### DIFF
--- a/downgrade
+++ b/downgrade
@@ -12,8 +12,8 @@ $(gettext "Options"):
   --pacman-conf   <$(gettext "path")>
                   $(gettext "pacman configuration file, defaults to") "/etc/pacman.conf"
   --pacman-cache  <$(gettext "path")>
-                  $(gettext "pacman cache directory or directories,")
-                  $(gettext "default value taken from pacman configuration file,")
+                  $(gettext "pacman cache directory,")
+                  $(gettext "default value(s) taken from pacman configuration file,")
                   $(gettext "or otherwise defaults to") "/var/cache/pacman/pkg"
   --pacman-log    <$(gettext "path")>
                   $(gettext "pacman log file,")
--- a/downgrade
+++ b/downgrade
@@ -168,11 +168,16 @@ search_packages() {
 
   if ((DOWNGRADE_FROM_CACHE)); then
     # Delay this defaulting so #read_pacman_conf behavior is tested
-    : "${PACMAN_CACHE:=$(read_pacman_conf CacheDir)}"
-    : "${PACMAN_CACHE:=/var/cache/pacman/pkg/}"
+    if ((!${#PACMAN_CACHE[@]})); then
+      mapfile -t PACMAN_CACHE < <(read_pacman_conf CacheDir)
+    fi
+
+    if ((!${#PACMAN_CACHE[@]})); then
+      PACMAN_CACHE=(/var/cache/pacman/pkg/)
+    fi
 
     # shellcheck disable=SC2086
-    find -L $PACMAN_CACHE -maxdepth "$DOWNGRADE_MAXDEPTH" -regextype posix-extended -regex ".*/$pkgfile_re"
+    find -L "${PACMAN_CACHE[@]}" -maxdepth "$DOWNGRADE_MAXDEPTH" -regextype posix-extended -regex ".*/$pkgfile_re"
   fi
 }
 
@@ -366,7 +371,7 @@ parse_options() {
       --pacman-cache)
         if [[ -n "$2" ]]; then
           shift
-          PACMAN_CACHE="$1"
+          PACMAN_CACHE+=("$1")
         else
           {
             gettext "Missing --pacman-cache argument"
--- a/README.md
+++ b/README.md
@@ -51,6 +51,12 @@ default behavior:
 $ downgrade foo bar
 ```
 
+Downgrade packages, specifying multiple cache directories:
+
+```
+$ downgrade --pacman-cache /path/to/cache --pacman-cache /path/to/other/cache foo bar
+```
+
 Downgrade a package with any of the following version-filtering operators `=`,
 `=~`, `<=`, `>=`, `<` and `>`:
 
--- a/README.md
+++ b/README.md
@@ -22,8 +22,8 @@ Options:
   --pacman-conf   <path>
                   pacman configuration file, defaults to "/etc/pacman.conf"
   --pacman-cache  <path>
-                  pacman cache directory or directories,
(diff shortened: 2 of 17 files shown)


downgrade
```diff
@@ -172,7 +172,8 @@ search_packages() {
     : "${PACMAN_CACHE:=/var/cache/pacman/pkg/}"
 
     # shellcheck disable=SC2086
-    find -L $PACMAN_CACHE -maxdepth "$DOWNGRADE_MAXDEPTH" -regextype posix-extended -regex ".*/$pkgfile_re"
+    # shellcheck disable=SC2068
+    find -L ${PACMAN_CACHE[@]} -maxdepth "$DOWNGRADE_MAXDEPTH" -regextype posix-extended -regex ".*/$pkgfile_re"
```
pbrisbin, line 176:
```suggestion
    find -L "${PACMAN_CACHE[@]}" -maxdepth "$DOWNGRADE_MAXDEPTH" -regextype posix-extended -regex ".*/$pkgfile_re"
```

Now that this is a true array and we're not relying on word-splitting, we can quote it normally and ShellCheck won't complain.

pbrisbin (dismissed):
LGTM.

Personally, I'd ignore the edge case. It a package appears twice it's not a huge deal IMO -- the user could select either and it would function. If someone finds this case and thinks it's confusing enough to report it to us, then we could consider addressing it.

downgrade
atreyasha: Yes true, but quoting leads to the following test error:

```
--- test/search_cache/pacman-conf-multiple.t
+++ test/search_cache/pacman-conf-multiple.t.err
@@ -13,7 +13,4 @@
   >   "CacheDir = $cache_1/" \
   >   "CacheDir = $cache_2/"
   > DOWNGRADE_FROM_CACHE=1 search_packages 'foo' | cut -d . -f 3- | sort
-  1/foo-1.0-1-any.pkg.tar.gz (glob)
-  1/foo-3.5-1-any.pkg.tar.xz (glob)
-  2/foo-1.1-1-any.pkg.tar.gz (glob)
-  2/foo-2.0-1-any.pkg.tar.gz (glob)
+  find: '/tmp/cramtests-b966w8q0/tmp/tmp.IJ46nCvHJo.1/\n/tmp/cramtests-b966w8q0/tmp/tmp.9OoZEEitNr.2/': No such file or directory
................
# Ran 34 tests, 0 skipped, 1 failed.
make: *** [Makefile:31: test] Error 1
```

Seems like quoting the array results in the `\n` character not being expanded as a new-line.

pbrisbin: Why is there a `\n`?

pbrisbin: I think I see...

Making `PACMAN_CACHE` an array when read from CLI should be matched by also making it an array when read from `pacman.conf`. Though it would function (with the quoting-hack), I wouldn't be comfortable with the variable being a different type depending on where it came from.

So instead of `PACMAN_CACHE="dir-one\ndir-two"` when read from `pacman.conf`, it needs to read that as `PACMAN_CACHE=( "dir-one" "dir-two" )`. `read -a` or `mapfile` I think are two ways to do that in Bash?

atreyasha: In this particular test case, two cache directories are read from a hypothetical `pacman.conf` file. These are separated by a newline, which is why it lands up there.

https://github.com/pbrisbin/downgrade/blob/6fe569bf82db27a0b8351fe708faf6785015c96a/downgrade#L37

pbrisbin: Yup, that will need changing/working-around.

> So instead of PACMAN_CACHE="dir-one\ndir-two" when read from pacman.conf, it needs to read that as PACMAN_CACHE=( "dir-one" "dir-two" ). read -a or mapfile I think are two ways to do that in Bash?

atreyasha: Hmm I am a bit lost how to make this work. So I see two approaches here.

1. We modify the `read_pacman_conf` function to print the matches on a single line separated by a space-character. Then this can be directly read in as an array.

2. Alternatively, we could modify how the default `PACMAN_CACHE` variable is assigned to an array. This is what we have right now:

https://github.com/pbrisbin/downgrade/blob/6fe569bf82db27a0b8351fe708faf6785015c96a/downgrade#L171-L172

I was thinking something like this below, but it reads the parentheses as literals. Same happens when I try adding a `read` command inside. Any suggestions?

```shell
 : "${PACMAN_CACHE:=($(read_pacman_conf CacheDir))}" 
```

atreyasha: If the bash parameter expansion syntax is too unfriendly for array assignment, we could replace lines 171 and 172 above with an "if-statement" where we check for an existing declaration of `PACMAN_CACHE`, if not we assign it the default values.

pbrisbin: Right, I was suggesting (2). Something like this should work:

```bash
if ((!{$#PACMAN_CACHE})); then
  mapfile -t PACMAN_CACHE < <(read_pacman_conf CacheDir)
fi

if ((!{$#PACMAN_CACHE})); then
  PACMAN_CACHE=( /var/cache/pacman/pkg/ )
fi
```

atreyasha: Yup this works great. Edited and adjusted the commit history.

pbrisbin (dismissed):
LGTM -- free free to merge after tackling those other book-keeping TODOs.

atreyasha:
Hmm strange GitHub used to allow me to commit after your reviews without blocking the merge. Anw now I need to request another review, sorry.

Did the routine book-keeping stuff. Also added a new example of `downgrade` to `README.md` which shows usage of multiple cache directories.

pbrisbin:
That seems to be why. I had this thing checked, which I've no un-checked:

pbrisbin (approved):

Issue #138, opened 2020-08-30
Configuration file for easier customization and modify downgrade execution workflow with assumption (and enforcement) of root permissions


atreyasha:
This issue proposes the introduction of user-specific configuration file(s) in order to customize the behaviour of `downgrade`. This is becoming increasingly important as `downgrade` incorporates more diverse features. Below are some points that could be included in a corresponding PR:

1. I would propose naming the configuration file as `downgrade.conf` and placing it in `~/.config/downgrade`.

2. For consistency and (prospective) easy debugging, we could deprecate the command-line input of environmental variables such as `DOWNGRADE_NOSUDO=1 downgrade` and instead fix the environmental variables as defaults within `downgrade` by replacing the `:=` variable assignment operator with `=`. The only way to change environmental variables inside `downgrade` would be via CLI options and/or a configuration file.

2. We set the assignment order of environmental variables to be -> first set them to defaults as per `downgrade`, then update them with user-specific configuration files, then update them by any CLI options supplied.

4. Include `man 5` entry for the syntax of this configuration file.

pbrisbin:
Sounds good to me.

I just had an idea that might simplify things: what if we only support altering downgrade behavior through command-line flags, and the configuration file can just pre-specify them? This is how (e.g.) [reflector](https://wiki.archlinux.org/index.php/Reflector#systemd_timer) works:

```
# /etc/xdg/reflector/reflector.conf
--save /etc/pacman.d/mirrorlist
--country France,Germany
--protocol https
--latest 5
```

I could see a very similar thing:

```
--pacman-conf /etc/foo/bar.conf
--nosudo
```

I also wonder if we should re-visit the idea of `downgrade` being invoked by a user and then internally using `sudo` (or `su root`) to execute root-level commands. It might be simpler for `downgrade` to assume a system-level governance (it's basically a `pacman` operation right?) and so assume it will itself be called by root. This would simplify away the `--nosudo` functionality nicely. If we did that, the configuration file should probably be system-wide, at `/etc/xdg/downgrade/downgrade.conf` -- again following Reflector's lead here. We could install a sample too.

atreyasha:
> I just had an idea that might simplify things: what if we only support altering downgrade behavior through command-line flags, and the configuration file can just pre-specify them?

:+1: for this, it is indeed much simpler and quicker to implement.

[...]

PR #144, opened 2020-10-25
Goodbye --nosudo

atreyasha:
### Checklist

- [x] Update cram unit tests
- [x] Update mandoc

### Description

This PR addresses part of #138 by removing `--nosudo` related logic. ~Will hold merge until the other part of #138 is complete, ie. fixing the configuration file workflow.~

### DIFF
--- a/downgrade
+++ b/downgrade
@@ -413,10 +413,6 @@ parse_options() {
         DOWNGRADE_FROM_ALA=0
         DOWNGRADE_FROM_CACHE=1
         ;;
-      -h | --help)
-        usage
-        exit 1
-        ;;
       --)
         shift
         pacman_args=("$@")
@@ -471,6 +467,14 @@ if ((!LIB)); then
   fi
 
   export TEXTDOMAIN=downgrade TEXTDOMAINDIR
+
+  # check for help CLI option
+  for arg; do
+    if [[ "$arg" =~ ^-h$|^--help$ ]]; then
+      usage
+      exit 0
+    fi
+  done
 
   # check to ensure downgrade running as root
   if ((UID)); then
--- a/downgrade
+++ b/downgrade
@@ -25,7 +25,6 @@ $(gettext "Options"):
                   $(gettext "location of ALA server, defaults to") "https://archive.archlinux.org"
   --ala-only      $(gettext "only use ALA server")
   --cached-only   $(gettext "only use cached packages")
-  --nosudo        $(gettext "do not use sudo even when available, use su")
   -h, --help      $(gettext "show help script")
 
 $(gettext "Note"):
--- a/downgrade
+++ b/downgrade
@@ -96,17 +96,17 @@ prompt_to_ignore() {
     if [[ "${ans,,}" == $(gettext 'y')* ]]; then
       ln="$(grep -n -m 1 '^ *IgnorePkg' "$PACMAN_CONF" | cut -d : -f 1)"
       if [ -n "$ln" ]; then
-        sudo sed -i "$ln s/.*/& $pkg/" "$PACMAN_CONF"
+        sed -i "$ln s/.*/& $pkg/" "$PACMAN_CONF"
         continue
       fi
 
       ln="$(grep -n '^# *IgnorePkg' "$PACMAN_CONF" | tail -n 1 | cut -d : -f 1)"
       if [ -n "$ln" ]; then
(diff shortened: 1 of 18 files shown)


atreyasha:
Just a quick question in regards to this modification. With the current diff, I noticed that now everything will require a `sudo` command for `downgrade`. So for example someone searching for help would input `downgrade --help` and the error message `downgrade must be run as root` will be brought up.

We could modify the parser to make things such that only the help option would be parsed without root permissions, but this would lead to a more complex diff. WDYT?

atreyasha:
Or perhaps a slightly repetitive option would be to add the following before calling `cli`:

```shell
if (($(id -u))); then
    {
      gettext "downgrade must be run as root"
      echo
    } >&2

    for arg; do
      if [ "$arg" == "--help" ] || [ "$arg" == "-h" ]; then
        usage
      fi
    done

    exit 1
fi
```

atreyasha:
Update on the last two comments: I moved the check for `-h|--help` earlier so that they are checked first before checking for root permissions. I found this better because also `pacman` does it this way, ie. it only requires root permissions when transacting package updates etc., but not for seeing the help script.

downgrade
```diff
@@ -96,17 +96,17 @@ prompt_to_ignore() {
...
 
       ln="$(grep -n '^# *IgnorePkg' "$PACMAN_CONF" | tail -n 1 | cut -d : -f 1)"
       if [ -n "$ln" ]; then
-        sudo sed -i "$ln s/.*/&\\nIgnorePkg = $pkg/" "$PACMAN_CONF"
+        sed -i "$ln s/.*/&\\nIgnorePkg = $pkg/" "$PACMAN_CONF"
         continue
       fi
 
-      printf "IgnorePkg = %s\\n" "$pkg" | sudo tee -a "$PACMAN_CONF"
+      printf "IgnorePkg = %s\\n" "$pkg" | tee -a "$PACMAN_CONF"
```
pbrisbin, line 109:
```suggestion
      printf "IgnorePkg = %s\\n" "$pkg" >> "$PACMAN_CONF"
```

I don't feel strongly, but this would work now since the shell itself has privileges now.

downgrade
```diff
@@ -484,5 +468,22 @@ if ((!LIB)); then
 
   export TEXTDOMAIN=downgrade TEXTDOMAINDIR
 
+  # check for help CLI option
+  for arg; do
+    if [[ "$arg" =~ ^-h$|^--help$ ]]; then
+      usage
+      exit 1
```
pbrisbin, line 475: Previous behavior, but I don't think `-h|--help` should exit unsuccessfully. The program did succeed in printing usage as asked.

Printing usage in response to an invalid argument should be non-zero, which is I think why this got this way by mistake.

downgrade
```diff
@@ -484,5 +468,22 @@ if ((!LIB)); then
...
+    fi
+  done
+
+  # check to ensure downgrade running as root
+  if (($(id -u))); then
```
pbrisbin, line 480:
```suggestion
  if ((UID)); then
```

Avoids spawning another process, and just a bit aesthetically nicer, IMO.

pbrisbin (approved):
Nice. Left some suggestions, but nothing blocking.

downgrade
atreyasha: Makes sense. Added all of the suggestions and cleaned the commit history.

downgrade
atreyasha: Ah yes, changed this.

downgrade
atreyasha: Will change.

downgrade
atreyasha: PS: Seems like this is a new GitHub feature that every new commit requires a new approval. Either that or I am somehow dismissing your reviews, which I don't know how haha.

In any case, I willl re-request your approval once I am done with the last step ie. locales.

pbrisbin: That's my fault,



I must not have hits **Save** when I un-checked this earlier, it was checked.

I made sure to hit Save now, so we should be good going forward. If not, there's a bug in GitHub here somewhere.

atreyasha:
FYI, replaced the `$` symbols in `README.md` to `#`; which AFAIK indicate root privileges.

pbrisbin:
Makes total sense. Good catch!

pbrisbin (approved):
Awesome! Major bump I assume?

atreyasha:
Nice :smile: I would suggest a minor bump for now and a major one after we complete the configuration file. WDYT?

PR #146, opened 2020-10-29
Book-keeping before release

atreyasha:
### Checklist

- [x] Update cram unit tests: irrelevant
- [x] Update `zsh`, `fish` and `bash` shell-completions: irrelevant
- [x] Update mandoc: irrelevant
- [x] Update usage function and readme: irrelevant

### Description

Just some book-keeping stuff before release:

1. Updated the changelog assuming a minor bump, but I can change that if we agree to something else.

2. Modified the version filtering example in the readme to have single quotes instead of double, since the double ones might lead to undesired variable expansions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
-## [Ongoing](https://github.com/pbrisbin/downgrade/compare/v8.1.1...master)
+## [Ongoing](https://github.com/pbrisbin/downgrade/compare/v9.0.0...master)
 
 None
+
+## [v9.0.0](https://github.com/pbrisbin/downgrade/compare/v8.1.1...v9.0.0)
+
+- [NEW] Now accepts multiple cache directory options
+- [NEW] Removal of --nosudo CLI option, downgrade now expects root privileges 
 
 ## [v8.1.1](https://github.com/pbrisbin/downgrade/compare/v8.1.0...v8.1.1)
 
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Downgrade a package with any of the following version-filtering operators `=`,
 `=~`, `<=`, `>=`, `<` and `>`:
 
 ```
-# downgrade "foo=1.0.0-1" "bar>=1.2.1-1" "baz=~^1.2"
+# downgrade 'foo=1.0.0-1' 'bar>=1.2.1-1' 'baz=~^1.2'
 ```
 
 Downgrade a package, looking in only local cache:


CHANGELOG.md
```diff
@@ -1,7 +1,12 @@
-## [Ongoing](https://github.com/pbrisbin/downgrade/compare/v8.1.1...master)
+## [Ongoing](https://github.com/pbrisbin/downgrade/compare/v8.2.0...master)
 
 None
 
+## [v8.2.0](https://github.com/pbrisbin/downgrade/compare/v8.1.1...v8.2.0)
+
+- [NEW] Pacman cache directory accumulation to support multiple cache locations
+- [NEW] Removal of --nosudo CLI option, downgrade now expects root privileges
```
pbrisbin, line 8: This feels like a big change, the entire interface is completely changed. I was thinking of `v9` if only to signal that. On the other hand, misuse is not harmful, the error is clear, and the behavior once executed has not changed in a meaningful way... So I'm fine with only a minor bump.

CHANGELOG.md
```diff
@@ -1,7 +1,12 @@
-## [Ongoing](https://github.com/pbrisbin/downgrade/compare/v8.1.1...master)
+## [Ongoing](https://github.com/pbrisbin/downgrade/compare/v8.2.0...master)
 
 None
 
+## [v8.2.0](https://github.com/pbrisbin/downgrade/compare/v8.1.1...v8.2.0)
+
+- [NEW] Pacman cache directory accumulation to support multiple cache locations
```
pbrisbin, line 7: Just noting here that when I read "cache directory accumulation" I keep thinking it means the cache is accumulating files, not that the options accumulate. Not sure if it's worth rewording in the CHANGELOG?

> - [NEW] Now accepts multiple cache directory options

?

pbrisbin (approved):

CHANGELOG.md
atreyasha: Hmm true, it is indeed a big change and a major bump would definitely indicate that. I had qualms about consecutive major bumps taking place in the last few releases but I guess the changes are what they are and the versioning should reflect that. Major bump it is :)

CHANGELOG.md
atreyasha: Aha yes. Will change.

atreyasha:
v9.0.0 released :birthday:

CHANGELOG.md
pbrisbin: FWIW, I treat Arch packages very differently than almost any other versioning situation. In a library, doing a major version update probably results in a large set of users that will refuse to move up and a maintenance headache if you've committed to supporting older versions. In Arch, everyone will just `-Syu` and it'll be fine. I still try to apply the same rules about indicating kinds of changes, but there's zero hesitation about large jumps in version.

PR #149, opened 2021-05-10
Minor fix for zsh completions

atreyasha:
### Checklist

- [x] Update cram unit tests
- [x] Update mandoc

### Description

Minor fix for `zsh` autocompletions. Previous completions used `pacman -Qsq` which only queries locally installed packages. This is now changed to `pacman -Ssq` to ensure that downgrade completes
packages which may not be installed on local system.


pbrisbin:
I also made that check not required anyway.

PR #151, opened 2021-05-14
System-level configuration file and minor menu modification

atreyasha:
### Checklist

- [x] Update cram unit tests
- [x] Update mandoc

### Description

This PR addresses #138, with the most significant changes summarized below:

1. System-level configuration file locally found in `./conf/downgrade.conf`
2. Workflow to read and parse the configuration file before command-line arguments
3. Update `downgrade` menu with cache directories of local packages to address any possible duplicates

A sample of the modified `downgrade` menu is shown below:

```
# downgrade picom

Available packages (community):

    1)  picom    7.4  1  remote
-   2)  picom    7.5  1  remote
-   3)  picom    7.5  2  remote
+   4)  picom    7.5  3  remote
-   5)  picom    8    1  remote
-   6)  picom    8    2  remote
-   7)  picom    8    3  remote
-   8)  picom    8    3  /var/cache/pacman/pkg
    9)  picom    8    4  remote
   10)  picom    8    5  remote
-  11)  picom    8.1  1  remote
-  12)  picom    8.1  1  /var/cache/pacman/pkg
   13)  picom    8.1  2  remote
   14)  picom    8.2  1  remote

select a package by number: 
```

Other changes include:

1. Removing old environmental variables
2. Removing the ability to define environmental variables outside of `downgrade`, such that the user can only influence behaviour through command-line options
3. Enforce certain arrays inside `downgrade` are unique to prevent double processing

### TODOs

* Update `Makefile` to install `downgrade.conf` in `/etc/xdg/downgrade/` if this is not present
* Update tests to reflect new features
* Standard book-keeping tasks


downgrade
```diff
@@ -35,6 +35,8 @@ EOF
 
 read_pacman_conf() { sed '/^#\?'"$1"' *= *\(.*\)$/!d; s//\1/' "$PACMAN_CONF"; }
 
+read_downgrade_conf() { grep -v '^\s*$\|^\s*\#' "$DOWNGRADE_ROOT_CONF" 2>/dev/null | xargs; }
```
pbrisbin, line 38: I came up with the following `grep` -- does it work? I find it easier to reason about, but it could be subjective.

```suggestion
read_downgrade_conf() { grep -E -v '^ *(#.*)?$')  "$DOWNGRADE_ROOT_CONF" 2>/dev/null | xargs; }
```

downgrade
```diff
@@ -448,12 +425,41 @@ parse_options() {
 }
 
 cli() {
-  parse_options "$@"
+  declare -a system_args
+
+  # get system and command-line arguments and parse everything
+  IFS=' ' read -r -a system_args <<<"$(read_downgrade_conf)"
```
pbrisbin, line 431: Can we add a test suite on this behavior? It's going to be tricky and I'd like to avoid regressions in the future.

I don't think we have any options that accept arguments with spaces in them right now, but if it's not too complicated, I think we should try to write this code to handle that now. In other words:

```conf
# Some header comment

# --example option
--actual "Option with spaces"
```

Should return `["--actual", "Option with spaces"]`. I think the code you have would return `["--actual", "Option", "with", "spaces"]`, right?

downgrade
```diff
@@ -448,12 +425,41 @@ parse_options() {
...
+  parse_options "${system_args[@]}" "$@"
+
+  # make arrays unique
+  IFS=' ' read -r -a terms <<<"$(printf "%s\n" "${terms[@]}" | uniq | tr '\n' ' ')"
+  IFS=' ' read -r -a PACMAN_CACHE <<<"$(printf "%s\n" "${PACMAN_CACHE[@]}" | uniq | tr '\n' ' ')"
```
pbrisbin, line 436:
```sh
read_unique() {
  local var=$1
  shift

  IFS=$'\n' read -r -a var <<<"$(printf "%s\n" "$@" | uniq)"
}
```

```suggestion
  read_unique terms "${terms[@]}"
  read_unique PACMAN_CACHE "${PACMAN_CACHE[@]}"
```

downgrade
```diff
@@ -448,12 +425,41 @@ parse_options() {
...
+
+  # proceed with rest of workflow
   main "${terms[@]}"
   pacman -U "${pacman_args[@]}" "${to_install[@]}"
   prompt_to_ignore "${to_ignore[@]}"
 }
 
+# Set script defaults
```
pbrisbin, line 444: You dropped the "Up-convert older variables", which is fine, but I think would mean a major version bump?

downgrade
```diff
@@ -448,12 +425,41 @@ parse_options() {
 }
 
 cli() {
-  parse_options "$@"
+  declare -a system_args
```
pbrisbin, line 428: Does `declare` within a function imply `local`? I have no idea. I would usually do `local system_args=()`.

I also wonder if `configuration_args` / `conf_args` is a better name. I'm not sure there's much meaning in "system" yet.

pbrisbin (changes requested):
Only blocking on tests, which I think are important to have for this.

atreyasha:
Nice. Yes I plan to tackle the tests once we are satisfied with everything else.

Btw what do you think of the new menu design? I removed the brackets to reduce clutter.

downgrade
atreyasha: This reads much better, thanks. There is a minor syntax error with an extra `)` next to the regular expression, but I will fix that with a rebase.

downgrade
atreyasha: Yes, since we are going away from user-inputted environmental variables altogether; I thought it would be good to remove these as well.

Major version bump sounds appropriate.

downgrade
atreyasha: Thanks :+1:

downgrade
atreyasha: Good catch! Yes this could be problematic and I will add a test suite for this later on.

Indeed, the code as of now will greedily split on spaces. Any suggestions btw on how to allow the splits to respect a clearly demarcated string? I will look into this further as well.

downgrade
atreyasha: I can't recall where I read this, but `declare` does imply locality.

But I agree with you, using the `local` syntax is more consistent with the coding style we have in `downgrade`. Will also change the variable name to `conf_args` :+1:

downgrade
atreyasha: Used `mapfile` instead because `read` was not able to parse newlines, not sure why. Also changed `uniq` to `sort -u` since `uniq` only considers adjacent dupicates and can therefore fail on several cases. This version works:

https://github.com/pbrisbin/downgrade/blob/e99d30aad608e9412e72ddffea2e792d92ef1b04/downgrade#L40-L46

downgrade
atreyasha: Used two tricks to solve this.

1. Replace `xargs` with `tr` to ensure any quotes or demarcations stay intact:

https://github.com/pbrisbin/downgrade/blob/e99d30aad608e9412e72ddffea2e792d92ef1b04/downgrade#L38

2. Use standard variable expansions to parse the single-line string extracted from the configuration file:

https://github.com/pbrisbin/downgrade/blob/e99d30aad608e9412e72ddffea2e792d92ef1b04/downgrade#L439

atreyasha: Let me know if these fixes make sense to you. The `eval` statement is a bit unusual but I can't think of anything else at the moment.

atreyasha: Added a test suite for this in 503b4389888b777d02c25f32c3934432fca6ca66

conf/downgrade.conf
```diff
@@ -0,0 +1,14 @@
+# /etc/xdg/downgrade/downgrade.conf
+#
+# Configuration file for downgrade
+# See downgrade(8) for more information on arguments
+# Uncomment and define relevant values
```
pbrisbin, line 5: Take it or leave it, but I might say,

```suggestion
# Defaults are shown, uncomment and define values as desired 
```

downgrade
```diff
@@ -179,7 +191,8 @@ search_packages() {
     fi
 
     # shellcheck disable=SC2086
-    find -L "${PACMAN_CACHE[@]}" -maxdepth "$DOWNGRADE_MAXDEPTH" -regextype posix-extended -regex ".*/$pkgfile_re"
+    find -L "${PACMAN_CACHE[@]}" -maxdepth "$DOWNGRADE_MAXDEPTH" \
+      -regextype posix-extended -regex ".*/$pkgfile_re" 2>/dev/null
```
pbrisbin, line 195: What prompted the `2>/dev/null`?

downgrade
```diff
@@ -210,7 +223,7 @@ output_package() {
 
   # Remote or local file
   if [[ $path =~ ^/ ]]; then
-    location="$(gettext 'local')"
+    location="$(dirname "$path")"
```
pbrisbin, line 226: :+1:

test/options_sanity/array_uniqueness.t
```diff
@@ -0,0 +1,15 @@
+  $ source "$TESTDIR/../helper.sh"
+
+Check that relevant arrays are made unique if they contain elements, otherwise they should be left as empty arrays
```
pbrisbin, line 3: I think it would be clearer to organize these tests such that each test case is only testing one thing, and that we do a suite by function. As they are, the unique assertions and parsing assertions are a bit mixed.

**test/parse_options.t**

```
Reads multiple --pacman-cache options into PACMAN_CACHE

  $ parse_options --pacman-cache /home/ --pacman-cache /away/ --pacman-cache /home/ foo 2>/dev/null
  > echo "${PACMAN_CACHE[@]}"
  /home/ /away/ /home/

Reads arguments into terms

  $ parse_options foo bar foo 2>/dev/null
  > echo "${terms[@]}";
  foo bar foo
```

**test/read_unique.t**

```
Sets unique arguments into a variable

  $ read_unique X x y y z y z
  > echo "${X[@]}"
  x y z
```

WDYT?

downgrade
```diff
@@ -448,12 +436,33 @@ parse_options() {
 }
 
 cli() {
-  parse_options "$@"
+  local conf_args=()
+
+  # Get configuration and command-line arguments and parse everything
+  eval "conf_args=($(read_downgrade_conf))"
```
pbrisbin, line 442: I agree this is probably the only way to do it. I wonder if there are any security concerns with this here.

pbrisbin (approved):

conf/downgrade.conf
atreyasha: Take it

downgrade
atreyasha: I wrongly typed a path in `PACMAN_CACHE` and that resulted in `find` throwing an error message. I thought it would be good to route this error message away, as we already did in other cases within `downgrade`. WDYT?

test/options_sanity/array_uniqueness.t
atreyasha: True, it makes sense to disentangle the functions.

Although I might add that `test/options_sanity/options_sanity.t` already tests option parsing similar to your first suggestion, albeit without duplicate packages/cache-directories.

In that case should we just replace `test/options_sanity/array_uniqueness.t` with your second suggestion?

downgrade
atreyasha: Hmm yes I think there is a security risk here, for example if your `conf` file has malicious code such as:

```
--pacman-cache "$(sudo rm -rf /)"
```

This line would actually get executed during the `eval` call, which makes me slightly uncomfortable.

How about an alternative where we use regular expressions to split all non-commented lines in the configuration file into one option and one value (if present)? Will try and provide a sample in a bit,

downgrade
pbrisbin: I try not to avoid suppressing errors unless there's a reason. If an end-user made a mistake in `--pacman-cache`, would it be better to show them what `find` said about it or hide it? I don't know the answer, but that's the question I'd ask.

> as we already did in other cases

This is definitely a case-by-case decision IMO, but if you describe the other cases, maybe there is a generally applicable way to decide for this one.

test/options_sanity/array_uniqueness.t
pbrisbin: > we just replace test/options_sanity/array_uniqueness.t with your second suggestion?

Sure, SGTM.

downgrade
pbrisbin: Happy to look at that code and see, but I wouldn't consider that possibility and immediate reason to change the implementation.

"The code in the conf will be `eval`-d" isn't itself a vulnerability. It requires that the code in the conf can somehow be attacker-controlled. And if an attacker can change `/etc/downgrade.conf` on you, I think you have bigger problems, right?

atreyasha: Hmm I am reading a lot of advice to avoid `eval` online, but you do have a point as well.

I tried the regex approach but it seems like an overkill. I found this one simple approach using `xargs` (source: https://superuser.com/a/1519494):

```shell
$ a='"aString that may haveSpaces IN IT" bar foo "bamboo" "bam boo"'
$ printf "%s" "$a" | xargs -n 1 printf "%s\n"
aString that may haveSpaces IN IT
bar
foo
bamboo
bam boo
```

I tested this locally and it does not execute any commands during the reading process. WDYT?

pbrisbin: I think we can consider the `eval` save by passing the elements through `printf '%q'` first. Sorry, I had used this trick before and since forgotten. It actually makes things simpler in general, I think (no need for `tr` or messing with `$IFS`):

```console
safe_example() {
> cat <<'EOM'
> --foo bar
> --baz
> --bat "a hoy"
> EOM
> }
```

```console
evil_example() {
cat <<'EOM'
> --foo
> $(ls /etc)
> --bar
> EOM
> }
```

```console
[patrick@leo ops]$ eval "x=( $(safe_example | xargs printf '%q ') )"
[patrick@leo ops]$ printf '>%s<\n' "${x[@]}"
>--foo<
>bar<
>--baz<
>--bat<
>a hoy<
```

```console
[patrick@leo ops]$ eval "y=( $(evil_example | xargs printf '%q ') )"
[patrick@leo ops]$ printf '>%s<\n' "${y[@]}"
>--foo<
>$(ls<
>/etc)<
>--bar<
```

atreyasha: Oh that's awesome! I will test this more and get back tomorrow. Wishing you a nice start of the week :)

downgrade
atreyasha: Yes true. The other case where I had used `2>/dev/null` was in the `grep` call for reading `downgrade.conf`.  But there it makes sense, because a user could opt not to have a `downgrade.conf` file and should not receive constant error messages about it being missing; since it is ultimately optional.

But I guess the argument is different for `find`, in that an error message could be useful there. In any case, I dropped this commit pertaining to `find` and left it as it was before.

atreyasha:
Addressed all the points, WDYT?

Btw am working on adding a line to the `install` PHONY such that the configuration file is copied over if it does not already exist. I believe the `install` commands overwrite everything by default. Do you know of an alternative implementation like `install` which does not force overwrite an already existing configuration file?

pbrisbin (approved):
Code LGTM.

Regarding installation: I think your options are to either use `mkdir -p` + `cp -n`, or to do your own `[ -f ... ]` check. Note that this won't matter to Arch users. The file will never be present in the fake-root that `make install` is run into to create the package, so that'll always happen. We will need to specify the file in the [`backup`](https://wiki.archlinux.org/title/PKGBUILD#backup) array so that on installation it ends up at `/etc/downgrade.conf.pacnew` if/when the user has made their own modifications. That said, we may still want `make install`, when run on a real system, to not clobber; though personally, I don't care much.

atreyasha:
I see. I am a bit confused because I have never worked with this `backup` array before. But in any case I checked the `PKGBUILD` of [`reflector`](https://github.com/archlinux/svntogit-community/blob/packages/reflector/trunk/PKGBUILD) and it appears they listed their configuration file both in the `install` command and in the `backup` array.

Going from this example, I assume we do the same and add the configuration file in the `Makefile`as per b4c41e9ba787407596ee984908c4efccb2590737 and also in our `PKGBUILD` with a new line containing `backup=(etc/xdg/downgrade/downgrade.conf)`.

Makefile
```diff
@@ -2,6 +2,7 @@ PREFIX       ?= /usr/local
 LOCALEPREFIX ?= $(PREFIX)/share/locale
 MANPREFIX    ?= $(PREFIX)/share/man
 PANDOC       ?= $(shell which pandoc)
+DOWNGRADE_CONF ?= /etc/xdg/downgrade/downgrade.conf
```
pbrisbin, line 5: I think this needs to respect `DESTDIR`. Otherwise, when used in a `PKGBUILD`, it would try to install onto the system _building_ the package, not the system _installing_ the package.

Makefile
```diff
@@ -33,6 +34,7 @@ test:
 .PHONY: install
 install:
 	install -Dm755 downgrade $(DESTDIR)$(PREFIX)/bin/downgrade
+	install -Dm644 conf/downgrade.conf $(DOWNGRADE_CONF)
 	install -Dm644 doc/downgrade.8 $(DESTDIR)$(MANPREFIX)/man8/downgrade.8
```
pbrisbin, line 38: I might follow these `PREFIX`/`LOCALEPREFIX`/`MANPREFIX` convention with something like:

```makefile
XDGPREFIX ?= /etc/xdg

install:
  install -Dm644 conf/downgrade.conf $(DESTDIR)$(XDGPREFIX)/downgrade/downgrade.conf
```

pbrisbin (changes requested):

Makefile
atreyasha: Done

atreyasha:
Let me know your thoughts on the lines I added to the `mandoc` and readme.

Not sure if you want to also add a separate `mandoc` for `downgrade.conf`, but seems like overkill since the configuration file is quite intuitive and has some instructions on it already.

Besides these two, I just need to do a book-keeping commit for the locales and then we can do a major version bump while manually modifying `PKGBUILD` with the new `backup` array.

doc/downgrade.8
```diff
@@ -20,13 +20,13 @@ output:
...
--  5)  terraform    0.12.0   1  (remote)
--  6)  terraform    0.12.0   1  (local)
-   7)  terraform    0.12.1   1  (remote)
+-  1)  terraform    0.11.11  2  remote
+-  2)  terraform    0.11.12  1  /var/cache/pacman/pkg
++  3)  terraform    0.11.13  1  remote
++  4)  terraform    0.11.13  1  /var/cache/pacman/pkg
+-  5)  terraform    0.12.0   1  remote
+-  6)  terraform    0.12.0   1  /var/cache/pacman/pkg
+   7)  terraform    0.12.1   1  remote
```
pbrisbin, line 29: :ok_hand:

doc/downgrade.8.md
```diff
@@ -53,10 +53,9 @@ The columns have the following meaning:
   The release of the package in cache or ALA.
 
 *location*
-  Possible values: {(remote)|(local)}
+  Possible values: {remote|/path/to/cache/dir}
 
-  If you have already downloaded this version, it will show *local*.\
-  *remote* indicates that the version is available on the ALA.
+  If you have already downloaded this version, it will show the cache directory where the package is located. *remote* indicates that the version is available on the ALA.
```
pbrisbin, line 58: Can you maintain the hard-wrapping at 80 for any .md files? I thought Restyled was set up here, and `prettier-markdown` should be doing that automatically :thinking:

pbrisbin (approved):
Awesome, thanks! This is such a great feature, I'm excited.

doc/downgrade.8.md
atreyasha: I can't recall exactly why I did not add all markdowns to `.restyled.yaml`. Will do it now and then take it from there.

atreyasha: Edited the markdowns and the `.restyled.yaml` file. For some reason the web-hook did not get triggered with a new commit. Is it possible for you to manually trigger a check?

pbrisbin: I wouldn't worry about it, could be any number of reasons for the hiccup.

pbrisbin:
FYI: restyled seems to be failing due to a syntax error in the configuration:

https://restyled.io/gh/pbrisbin/repos/downgrade/jobs

It's not reporting any status back to the PR because it needs to at least read the `restyled.yaml` to even know if it should be sending PR statuses. I could probably change that :thinking:

atreyasha:
Ah I see. Maybe I need to quote the asterisk. Let's try again...

atreyasha:
It works, thanks! Will fix the styling now.

atreyasha:
Okay I believe I fixed everything now. Let me know if you spot anything else, if not I will push the locale updates and release tomorrow with a major version bump.

pbrisbin:
My Approval from earlier still stands :smile:

atreyasha:
Released v10.0.0 :heart_eyes:

> Awesome, thanks! This is such a great feature, I'm excited.

Very happy with this feature too! Also thanks for the `aur-release` script, it proved super helpful while releasing especially with the `-e` flag which was needed to add the backup array.

Issue #147, opened 2020-11-02
Implement independent --version


atreyasha:
> What if we didn't build `--version` at all, and just had the Issue Template ask for `pacman -Qi downgrade`?
>
> Using `pacman`'s version comes with the downside that you might be running a `dowgrade` installed some other way. I'm OK with that downside, but probably only if we didn't implement our own `--version` that tried to do it. If we just asked users to supply `pacman`'s version information, they can say if they're installed in some other way. If we build it as `--version` the user wouldn't know that's happening and we might be supplying the wrong information, or throwing an error, etc, etc.
>
> So IMO we either build a real `--version` or we just ask the users to give us `pacman` details. I don't feel strongly which though.
>
> _Originally posted by @pbrisbin in https://github.com/pbrisbin/downgrade/issues/137#issuecomment-696210727_

Was thinking again about this issue and had a change of mind, I would like at some point in time to implement an independent `--version` CLI option outside of `pacman`.

In terms of implementation, I was thinking that we could add a line or two in the `Makefile` where on each new release; the new version will be appended somewhere in the `downgrade` script and perhaps also in the footer of the mandoc. Then any calls with `--version` would simply search for this version watermark and output it.

WDYT? Any suggestions on the exact implementation?

**Edit:** One problem with my suggestion here is multiple merges, since any commits from me to `downgrade` (eg. adding a version watermark before release) would require another approval and merge. To circumvent this, how about adding another `PHONY` to the `Makefile` (eg. called version-mark) which simply adds the version watermarks to relevant files based on the version bump we agree on. This could then be part of the book-keeping procedures before releasing. That way, we can leave the actual `make release*` just as it is; and all the commits to `downgrade` can happen before as part of book-keeping.

pbrisbin:
> WDYT? Any suggestions on the exact implementation?

If I understand it correctly, a common UNIX-ey way to do this would be to add `VERSION = ...` to `Makefile`, make any files that need to know about it `.in` files and then pre-process them with `m4`:

```makefile
VERSION = 1.0.0

downgrade: downgrade.in
  m4 -D @VERSION@=$(VERSION) $^ > $@
```

```sh
# downgrade.in

# ...

version() {
  echo "downgrade v@VERSION@"
}

# ...
```

You could get fancier, encoding git information and all that as well, but I defer to you on if we should explore that.

PR #156, opened 2021-05-20
Add --version command-line option

atreyasha:
### Checklist

- [x] Update cram unit tests
- [x] Update mandoc

### Description

This PR addresses #147 by proposing a new `--version` command-line option. This prints out the `DOWNGRADE_VERSION` variable that is defined internally.

I chose to forego the whole process with `m4` as it seemed bulky compared to the simple regular-expression based replacement that is required for this. Also I am sure we will never define the `DOWNGRADE_VERSION` variable in any other context in `downgrade`, so the absolute replacement using `sed` should be safe.

WDYT?

### DIFF
--- a/downgrade
+++ b/downgrade
@@ -486,6 +486,9 @@ if ((!LIB)); then
     if [[ "$arg" =~ ^-h$|^--help$ ]]; then
       usage
       exit 0
+    elif [[ "$arg" == '--version' ]]; then
+      printf "%s\n" "$DOWNGRADE_VERSION"
+      exit 0
     fi
   done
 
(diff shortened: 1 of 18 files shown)


atreyasha:
Oh btw one more thing, could you update the default branch from `master` to `main` on [codeclimate](https://codeclimate.com/github/pbrisbin/downgrade)?

I think it did not get updated, which is why those tests are not running.

downgrade
```diff
@@ -486,6 +487,9 @@ if ((!LIB)); then
     if [[ "$arg" =~ ^-h$|^--help$ ]]; then
       usage
       exit 0
+    elif [[ "$arg" =~ ^--version$ ]]; then
```
pbrisbin, line 490:
```suggestion
    elif [[ "$arg" == '--version' ]]; then
```

pbrisbin (approved):
SGTM!

I updated CC and re-installed the PR Integration and Webhook for good measure.

Issue #158, opened 2021-06-23
Add downgrade --unignore


kostadinnm:
### Background

Hello.
I'd like to propose an enhancement, which should be quite easy to implement, I guess.
I've downgraded a dependency of a package that refused to upgrade. Now, a new version of my main package is released, but I need to use other tool(`pacman`) to revert the effects of the downgrade. I'd memoize easier, if utilizing the same command interface for downgrading and reverting the downgrade actions.

### Proposed feature
`downgrade` should be able to revert package version freeze.

atreyasha:
Thanks for this describing this feature request.

If I understood you correctly, you propose a new feature where `downgrade` reverts its last action and restores a downgraded package back to its original version before the downgrade. Is this correct? If I got it wrong, it would be good if you could provide a specific example with mock packages/versions.

As of now, the only way of tracking downgraded packages would be to look at the `IgnorePkg` definition in `/etc/pacman.conf` and assume all packages there have been explicitly downgraded. Because no information on the pre-downgrade package version is tracked, it would not be possible to revert the action of downgrade back to the (specific) old version.

However, I could imagine the following new CLI option:

```
# downgrade --revert <pkg>
```

This would trigger the following workflow:

1. Check if the package is defined in `IgnorePkg`
2. If it is, remove its definition from `IgnorePkg`
3. Then execute `pacman -S <pkg>` to update the package to the latest version upstream

@kostadinnm would this workaround still be useful for your use-case?

@pbrisbin WDYT? I could imagine this feature request being a part of #118 since it also involves non-interactive actions similar to the ones mentioned in the issue.

pbrisbin:
Well, you know me, I'm always trying to find ways for `downgrade` to be useful _with_ other tools, instead of taking over their tasks...

What do you think of

```console
downgrade --unignore [pkg...]
```

The `--unignore` flag causes `downgrade` to remove anything present in `IgnorePkg`. The `pkg` argument is optional, can be specified multiple times, and filters what it will actually remove. It then prints, one per line, the packages that were affected.

Because of the printing, you could do:

```console
downgrade --unignore | pacman -S -
```

WDYT?

---

To go **fully** good-oss-citizen, we could consider extracting the parts of `downgrade` that interact with `IgnorePkg` to a separate executable, something like

```console
pacignore ls|add|rm [pkg...]
```

and see if we can get that added to `pacman-contrib`. Then `downgrade` could just use it, and this specific feature would end up being provided by that instead:

[...]

PR #159, opened 2021-06-24
Add `==` as a version operator alias of `=`

atreyasha:
### Checklist

- [x] No duplicate issues/PRs
- [x] Update usage function
- [x] Update unit tests
- [x] Update mandoc
- [x] Update readme
- [x] Update locales

### Description

This PR adds the `==` version operator as an alias of the already existing `=` operator. The reason why I added the `=` operator in the first place was to be consistent with the version comparison style in `pacman`. But then I realized that that some users might be more familiar with using `==` instead, so why not have them both?

**PS:** Let's release after this PR, I think we have enough new stuff for a minor version bump. Then I can move on to #118 and #158 after.


atreyasha:
Miscellaneous changes:

1. I found that the path of our new configuration file was wrong in the readme and mandoc, so I fixed these.
2. The changelog's `Ongoing` entry compared against the legacy `master` branch, so I fixed this to `main`.

pbrisbin (approved):

atreyasha:
@pbrisbin BTW could you allow me to have write access to `main`?

Our previous change to `Makefile` involves a final commit to `main` which updates the `DOWNGRADE_VERSION` variable inside `downgrade`, and I am currently not able to do this without write-access.

pbrisbin:
I don't think it's actually permissions (well not directly) that is blocking your push to `main`; I think it's branch protection, which requires status checks (and so a PR) for any non-admin pushing to that branch. I also can't seem to find anyway to make you an admin of this repository (we'd have to make an organization and move the repository to do that).

I'm happy to explore the organization approach, but in the meantime why can't we do the version change in a PR?

atreyasha:
> but in the meantime why can't we do the version change in a PR?

That's possible as well, just more repetition of things that can automated. Will open one now.

pbrisbin:
> more repetition of things that can automated

On that topic, we should definitely look into GitHub Actions. I've been moving all my repositories to it, and it's easier to automate release activities there than Circle. I'll open an Issue and start discussing options.

PR #161, opened 2021-06-24
Release v10.1.0

atreyasha:
### Checklist

- [x] No duplicate issues/PRs
- [x] Update usage function
- [x] Update unit tests
- [x] Update mandoc
- [x] Update readme
- [x] Update locales

### Description

Changes are as follows:

1. Remove feature in `Makefile` which updates `DOWNGRADE_VERSION` automatically, since this results in branch protection problems
2. Add the task of updating `DOWNGRADE_VERSION` manually to GitHub PR templates
3. Update `DOWNGRADE_VERSION` to `v10.1.0` for this minor version bump and release

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,9 +72,6 @@ AUR_RELEASE_OPTIONS ?=
 .PHONY: release
 release: test
 	[ -n "$(VERSION)" ]
-	sed -E "s/^(DOWNGRADE_VERSION=)(.*)/\1\"v$(VERSION)\"/g" -i downgrade
-	git add downgrade
-	git commit -m "Release v$(VERSION)"
 	git tag -s -m v$(VERSION) v$(VERSION)
 	git push --follow-tags
 	aur-release $(AUR_RELEASE_OPTIONS) downgrade "$(VERSION)"
--- a/Makefile
+++ b/Makefile
@@ -55,18 +55,7 @@ uninstall:
 	  $(DESTDIR)$(PREFIX)/share/fish/vendor_completions.d/downgrade.fish \
 	  $(DESTDIR)$(LOCALEPREFIX)/*/LC_MESSAGES/downgrade.mo
 
-.PHONY: release.major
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -72,9 +72,6 @@ AUR_RELEASE_OPTIONS ?=
 .PHONY: release
 release: test
 	[ -n "$(VERSION)" ]
-	sed -E "s/^(DOWNGRADE_VERSION=)(.*)/\1\"v$(VERSION)\"/g" -i downgrade
-	git add downgrade
-	git commit -m "Release v$(VERSION)"
```
pbrisbin, line 77: Oh, I see how this works now. I'd recommend we remove the need to specify `VERSION` for `make`, and use something like:

```make
VERSION ?= $(shell sed -E /^DOWNGRADE_VERSION=(.*)/!d; s//\1/' downgrade)
```

That way, the version is in exactly one place (the script).

pbrisbin (approved):

Makefile
atreyasha: That works, but it carries the quotes as well. How about this?

```shell
sed -E '/^DOWNGRADE_VERSION=(.*)/!d; s//\1/' downgrade | tr -d "'" | tr -d '"'
```

atreyasha: I tried removing the quotes within `sed` but I keep getting syntax errors because of nested double and single quotes. This option worked easily.

pbrisbin: Well I think we have tight control on this variable, so we can either not quote it, or quote it consistently and expect that in the regex, no need to be flexible withe single-double, `tr`.

```sh
sed '^DOWNGRADE_VERSION="\([^"]*\)".*$/!d; s//\1/'
```

In case you didn't realize, using `\(` and `\)` in this regex means you won't need `-E`. Since it's such a minor thing, I tend to do that to avoid the GNU-only flag.

atreyasha: Makes sense, thanks

PR #163, opened 2021-06-24
Patch downgrade

atreyasha:
### Checklist

- [x] No duplicate issues/PRs
- [x] Update usage function
- [x] Update unit tests
- [x] Update mandoc
- [x] Update readme
- [x] Update locales

### Description

Sorry, forgot that the `sed` command should also remove the `v` before the version.

### DIFF
--- a/downgrade
+++ b/downgrade
@@ -468,7 +468,7 @@ DOWNGRADE_FROM_ALA=1
 DOWNGRADE_FROM_CACHE=1
 DOWNGRADE_MAXDEPTH=1
 DOWNGRADE_CONF="/etc/xdg/downgrade/downgrade.conf"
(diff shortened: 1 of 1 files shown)


pbrisbin (approved):
Trimming the `v` or not is up to you. You could also _not_ add the `v` in L64. `aur-release` will work with or without it, IIRC.

Personally, if pressed, I would say that `x.y.z` is the version (no `v`) and a `v` prefix is added for the tag only, to denote that it's a "version tag for x.y.z". Under this thinking you shouldn't have a v in the in-script value, and you should add it only in the `git-tag` command below.

atreyasha:
I don't have a strong preference, will go with your suggestion.

atreyasha:
Released `v10.1.0` :heart:

Issue #168, opened 2021-12-31
Request to add 32 bit support


Practicalbutterfly5:
### Checklist

- [x] No duplicate issues/PRs
- [x] Are you running the latest `downgrade` release from the [AUR](https://aur.archlinux.org/packages/downgrade/)?

### Background

The current version of `downgrade` only looks in https://archive.archlinux.org and not https://archive.archlinux32.org, as a result, it doesn't downgrade/install packages on a 32-bit system. Using --ala-url as https://archive.archlinux32.org results in `Unable to downgrade` 
### Proposed feature
Requested to use https://archive.archlinux32.org for 32-bit systems.

atreyasha:
Hi @Practicalbutterfly5, thanks for opening this feature request.

Before #106, we actually did offer this option where the user could specify an alternative architecture with eg. the following CLI argument: `--arch i686`. However, we removed this option since Arch Linux no longer (officially) supports 32-bit systems. I was also not aware of the ongoing Arch Linux 32 [project](https://archlinux32.org/) before this.

In terms of diffs required, I would say it is fairly straightforward to revert (plus refine) this feature. We would have to reintroduce the `--arch` CLI option and fine-tune the package selection menu with the relevant architectures.

In terms of philosophy, I think it fits in with `downgrade` being a general purpose tool that can serve various user groups out-of-the-box. Here it is very useful that the ALA directory structures are basically the same across both projects.

@pbrisbin WDYT?

pbrisbin:
In an ideal word, IMO, `--ala-url https://archive.archlinux32.org` should Just Work, with no further changes to `downgrade`. So, I'd like to understand better why it does not and see if we could reasonable treat this as a bug-fix.

pbrisbin:
As far as I can tell, this minor change makes it all work just fine:

```diff
 # Set script defaults
 PACMAN="pacman"
 PACMAN_CONF="/etc/pacman.conf"
-DOWNGRADE_ARCH="x86_64"
+DOWNGRADE_ARCH="$(uname -m)"
 DOWNGRADE_ALA_URL="https://archive.archlinux.org"
 DOWNGRADE_FROM_ALA=1
 DOWNGRADE_FROM_CACHE=1
```

If I force it to `i686`, this works:

```console
# ./downgrade --ala-url https://archive.archlinux32.org --ala-only l-smash
Available packages (extra):

   1)  l-smash    2.9.1     1  remote
   2)  l-smash    2.14.5  1.0  remote
[...]

PR #172, opened 2022-01-08
Resolve `DOWNGRADE_ARCH` variable automatically

atreyasha:
### Checklist

- [x] No duplicate issues/PRs
- [x] Update usage function
- [x] Update unit tests
- [x] Update mandoc
- [x] Update readme
- [x] Update locales

### Description

This PR addresses #168

@pbrisbin should we release as a patch ie. `10.1.0 -> 10.1.1`?


pbrisbin (approved):
Change looks good, but I suspect you'll need to spoof this command for the tests?

atreyasha:
Yes, I spoofed the `pacman-conf` to just return an empty string

pbrisbin:
If it works 👍. You could also drop a mock (like we have for curl and pacman) in test/bin and record a fixture of it from your system.

pbrisbin:
Might be ideal to have a test case for the 32 but archive, but I'm not sure how involved that would be.

atreyasha:
> You could also drop a mock (like we have for curl and pacman) in test/bin and record a fixture of it from your system

Ah of course! That's sounds like a better option

> Might be ideal to have a test case for the 32 bit archive

Thought about it and was on the fence a bit. But yeah, will add a test case for posterity

atreyasha:
I only updated the `test/search_ala/ala.t` test to search for packages based on various architectures and ALAs. I think updating other tests might get more involved. WDYT?

atreyasha:
Also one more thing, I decided to patch the `release` target in `Makefile` with a check to ensure that we do not mistakenly release with the same version number. I mean this should never happen but I thought it is good to be safe.

atreyasha:
@pbrisbin just updating that I will merge and release later on today, in case you want to suggest something else for the test or `Makefile`

Makefile
```diff
@@ -61,6 +61,7 @@ AUR_RELEASE_OPTIONS ?=
 .PHONY: release
 release: test
 	[ -n "$(VERSION)" ]
+	[ "v$(VERSION)" != "$(shell git tag -l --sort -version:refname | head -n 1)" ]
 	git tag -s -m v$(VERSION) v$(VERSION)
```
pbrisbin, line 65: There should be no reason to explicitly check because `git tag` will fail (with an informative error) if you try to create a tag that already exists. That's even better than your check because it ensures not only that you don't re-release with the latest version again but you don't re-release with _any_ known version again.

pbrisbin (approved):

Makefile
atreyasha: Ah right, make sense. Will remove this commit.

Issue #169, opened 2022-01-04
Add more descriptive error messages


atreyasha:
### Checklist

- [x] No duplicate issues/PRs
- [x] Are you running the latest `downgrade` release from the [AUR](https://aur.archlinux.org/packages/downgrade/)?

### Background

As mentioned in https://github.com/pbrisbin/downgrade/issues/168#issuecomment-1004352850, `downgrade` only provides an uninformative error message `Unable to downgrade <pkg>`.

### Proposed feature

We should provide more descriptive error messages, such as no matching packages found, no results found, etc.

Issue #166, opened 2021-07-10
TUI


atreyasha:
### Checklist

- [x] No duplicate issues/PRs
- [x] Are you running the latest `downgrade` release from the [AUR](https://aur.archlinux.org/packages/downgrade/)?

### Proposed feature

Personally, I always pictured piping downgrade into something like `fzf` as the next iteration of interface:

```
downgrade search firefox | fzf | downgrade
```

_Originally posted by @pbrisbin in https://github.com/pbrisbin/downgrade/issues/164#issuecomment-877215106_

PR #174, opened 2022-01-15
TUI with `fzf`

atreyasha:
### Checklist

- [x] No duplicate issues/PRs
- [x] Update usage function
- [x] Update unit tests
- [x] Update mandoc
- [x] Update readme
- [x] Update locales

### Description

This PR addresses two issues:

1. We use `fzf` as a new interface for `downgrade` as per #166
2. We have a more descriptive error message as per #169

### Improvements

The test suites `main/invalid-choice.t` and `main/success.t` fail now particularly because of `fzf` getting in between the pipe: https://github.com/pbrisbin/downgrade/blob/6c0a18c987a1002d8dca687154a95c1747db4668/downgrade#L287-L289

I am not quite sure how to patch this, and currently I just spoofed `fzf` to `cat` such that the tests at least go to completion (they would previously hang waiting for `fzf`).

@pbrisbin any suggestions on how to fix this part? Also of course your thoughts on the new interface :smile:


pbrisbin (approved):
Neat! I have just a few thoughts:

1. Can you upload a gif of using it?
2. Could using `fzf` allow us to simplify (or ideally remove) `present_packages`? Presentation-related complexity would be something I thought we might be able to drop by using an external picker
3. For the tests, how about writing your test-helper `fzf` to look for an environment variable that tells is which item to "pick"?

   ```sh
   fzf() { sed -n "${CHOICE}p"; }
   ```

   This won't work for `-1`, but I don't think we need that test. Downgrades interface is to send lines to `fzf` and get lines back (or not), which `sed` can mimic exactly. Since users don't enter a number any more, we wouldn't have to test such a case.

atreyasha:
@pbrisbin thanks for the points. Will get to them all progressively, but for now here is the GIF for point 1:

demo

pbrisbin:
Wow, that's pretty cool!

So a little more about what I was thinking with (2): what if we output (to `fzf`) `{indicator} {url or path}`?

```
  + /var/cache/pacman/emacs-27.1.3-x86_64.pkg.xz
>   https://archive.archlinux.org/packages/e/emacs-26.1-x86_64.pkg.xz
    ...
    ...
Available packages (extra):
10/10
>
```

This does two things:

1. Usage with `fzf` is actually more powerful since a user could type (e.g.) `ca/27` to immediately filter to the local v27 line, or `ar/` to get to the remote line. Since this will include full versions and architecture (if that mattered) you could include that in your `fzf` expression.  IIUC correctly, `fzf` supports stuff like `^h` too, to more directly filter on this more "transparent" output.
2. Makes what we get back from `fzf` more useful; we only need to `cut -d' ' -f 2` to get `{url or path}`, download it if it's a `url`, then install it.

I get that this might sacrifice clarity to the user for the simplicity of working with `{url or path}` items, so this isn't a strong suggestion; just an option to consider.

atreyasha:
Regarding point 2 and your suggestion:

> Could using fzf allow us to simplify (or ideally remove) present_packages?

AFAIK, unfortunately this is not possible with `fzf` alone (by design apparently). As per https://github.com/junegunn/fzf/issues/1760, `fzf` expects formatting/presentation to be handled by separate programs.

> So a little more about what I was thinking with (2): what if we output (to fzf) {indicator} {url or path}?

I think this definitely will make `downgrade` simpler by reducing the need for explicit package presentation. But I agree that clarity (or structured viewing) from the perspective of the user will be lost. I am not sure if that might be the best option at the moment.

How about we keep the current presentation style for this PR, and we discuss the alternative presentation as a separate issue down the line?

pbrisbin:
> How about we keep the current presentation style for this PR, and we discuss the alternative presentation as a separate issue down the line?

Sure, SGTM.

atreyasha:
Regarding point 3, that works perfectly. Instead of `sed` I used `grep` to get the matching line:

https://github.com/pbrisbin/downgrade/blob/08eff09280e1cd85d9194a04ba975afa476326d7/test/helper.sh#L22

atreyasha:
Should we do a major/minor release after merging this PR?

In any case, I will update the `PKGBUILD` with the `fzf` dependency.

pbrisbin:
> Should we do a major/minor release

Good question. I defer to you, technically it need not be major since the ultimately interface hasn't _really_ changed in a breaking way. But it is a "major" change that could be worth signalling anyway.

atreyasha:
Will go with the major version update

atreyasha:
v11.0.0 released :heart:

PR #178, opened 2022-01-25
Replace Code Climate with GitHub action for shellcheck

atreyasha:
### Checklist

- [x] No duplicate issues/PRs
- [x] Update usage function
- [x] Update unit tests
- [x] Update mandoc
- [x] Update readme
- [x] Update locales

### Description

Remove `.codeclimate.yml` since we no longer use Code Climate for our checks.


pbrisbin (approved):
Hmm, I never intended to stop using Code Climate -- though it does frequently need to have the webhook re-installed. That said, Restyled's prettier-markdown should solve whatever markdownlint was doing, and we can add a new [GitHub Action for ShellCheck](https://github.com/ludeeus/action-shellcheck), and the webhook re-installation is indeed pretty annoying -- so I'm good with this :+1:.

:+1:

atreyasha:
Ah I see. I updated GitHub Actions with `shellcheck.yml` from [here](https://github.com/ludeeus/action-shellcheck).

atreyasha:
@pbrisbin is the current configuration alright? Or would it be better if I added the `shellcheck` job into `ci.yml` itself?

pbrisbin:
Good question. I'm not actually sure if there's a meaningful trade-off between distinct Workflows or multiple Jobs. I'm typically add Jobs, but I'm down to try a separate Workflow.

Do you mind naming it `lint.yml` / `name: Lint`? Just in case we ever grow into other "lint" activities.

atreyasha:
> Good question. I'm not actually sure if there's a meaningful trade-off between distinct Workflows or multiple Jobs. I'm typically add Jobs, but I'm down to try a separate Workflow.

I don't have much experience with GitHub Actions, so will defer to your opinion on this.

> Do you mind naming it lint.yml / name: Lint? Just in case we ever grow into other "lint" activities.

Done

pbrisbin (approved):

Issue #184, opened 2022-02-26
Packages with a slash in the name cannot be downgraded


nailuj29:
### Checklist

- [x] No duplicate issues/PRs
- [x] Are you running the latest `downgrade` release from the [AUR](https://aur.archlinux.org/packages/downgrade/)?

### Environment

* System information:

Linux nailuj29 5.15.21-1-MANJARO #1 SMP PREEMPT Sun Feb 6 12:21:42 UTC 2022 x86_64 GNU/Linux
* Downgrade version:

```
Downgrading from A.L.A. is disabled on the stable branch. To override this behavior, set DOWNGRADE_FROM_ALA to 1 .
See https://wiki.manjaro.org/index.php/Downgrading_packages  for more details.
```
11.0.0

### Description

Package downgrade fails when package name has a slash
### Steps to Reproduce

1. First Step
`downgrade 'extra/boost-libs'`

### Expected behavior

The package is downgraded
### Actual behavior
```
sed: -e expression #1, char 40: Unmatched ( or \(
No results found
Unable to downgrade extra/boost-libs
```

atreyasha:
Hi @nailuj29. In regards to downgrading `boost-libs`, just downgrading with the package name should do the trick. In other words, specifying the repository `extra` here is unnecessary:

```
# downgrade boost-libs
```

Also package names containing slashes are not permitted as per the Arch package [guidelines](https://wiki.archlinux.org/title/Arch_package_guidelines), so we would not need to account for this.

@pbrisbin on the other hand, I think providing a slash (or any other disallowed character) in the package name should ideally lead to an error message and not the `sed` error message `sed: -e expression ...`. I think this might have to do with these special characters not being escapred properly before the regular expression. Do you think it makes sense to escape the package entries beforehand with something like `printf "%q"`?

pbrisbin:
> I think providing a slash (or any other disallowed character) in the package name should ideally lead to a standard error message and not...

Agreed with the principle. In this case, it could be better to respond to the failure instead of aiming to prevent it, something like:

```sh
if ! { whatever contains the sed }; then
  echo "{better error message}. $input might be invalid?"
  exit 1
fi
```

Or, if `/` is by far the most common mistake, and it's a problem because we do `s/.../`, we could do anything else, such as `s|...|` or `s%...%`.

nailuj29:
@atreyasha 
When I run `downgrade boost-libs` it outputs this and fails.
```

[...]

PR #190, opened 2022-03-05
Bugfix for `sed` error messages

atreyasha:
### Checklist

* Admin:
  - [x] No duplicate PRs
* Integration:
  - [x] Update unit tests
  - [x] Update configuration file
* Documentation:
  - [x] Update usage function
  - [x] Update mandoc
  - [x] Update readme
  - [x] Update locales
* Release:
  - [x] Update change log
  - [x] Update `DOWNGRADE_VERSION`

### Description

This PR addresses #184 by directing `stderr` from all `sed` commands to `/dev/null`.

I considered wrapping the blocks/pipes  where `sed` is used with an if-statement that checks for their exit codes as per https://github.com/pbrisbin/downgrade/issues/184#issuecomment-1053652095. But given there are several such blocks with `sed`, I thought it might be simpler to just direct the error message to `/dev/null`. The downside is that we don't get a special error message specifying invalid input, but we would still get the `No results found` error message which is not incorrect either.

WDYT?


pbrisbin (commented):
I guess I'm just not clear on the behavior we're trying to fix.

If the behavior is

```console
% downgrade whatever
sed error...
[exit 1]
```

Then I don't like this fix. It'll just remove output and add nothing useful.

I think you're suggesting the current behavior is:

```console
% downgrade whatever
sed error...
No results found
[exit 1]
```

In which case, I can see the reasoning to just silence it.

However, my suggestion for silencing that sed error was built on the assumption that it was the only (-ish) one that led to the no-results outcome. Silencing a dozen `sed` errors seems like a recipe for leaving no indication what (or if anything) went wrong in a no-results case.

I think we have to look the `sed`s individually. If a given `sed` will "only" (with reasonable confidence) fail because the user's search term was invalid, then silencing it and letting "No results found" show is reasonable. Ideally, you'd see "invalid search", but I can get behind the simplification that an invalid search leads to no results.

But many (most?) of these `sed` lines are (e.g.) reading `pacman.conf`, parsing a package path, etc. Failures here would (probably?) be coding errors on our part. A blanket "No results found" is a bit of a problem.

So I would suggest that we only change the `sed` that we know failed in that bug report, the one that we know fails because of an invalid search term. I think that would be the most minimal bugfix and a net improvement.

As more comprehensive alternative, we could introduce something like this:

```sh
sed_msg() {
  local msg=$1
  shift

  if ! sed "$@"; then
    echo "Failed $msg" >&2
    exit 1
  fi
}
```

And then,

```diff
-sed -i "$ln s/.*/&\\nIgnorePkg = $pkg/" "$PACMAN_CONF"
+sed_msg "adding IgnorePkg" -i "$ln s/.*/&\\nIgnorePkg = $pkg/" "$PACMAN_CONF"
```

Or something...

atreyasha:
> I guess I'm just not clear on the behavior we're trying to fix.
> I think you're suggesting the current behavior is:
>
> ```
> % downgrade whatever
> sed error...
> No results found
> [exit 1]
> ```

Exactly, this is the current behaviour and I would like to ideally silence the `sed` error messages in such cases.

> So I would suggest that we only change the sed that we know failed in that bug report, the one that we know fails because of an invalid search term. I think that would be the most minimal bugfix and a net improvement.

True. I looked deeper and found that the `sed` errors arise from the `process_term` function:

https://github.com/pbrisbin/downgrade/blob/4c3af2cfee79f2c4265abe69961fa1cc2dca2f4b/downgrade#L322-L326

Somehow despite the inner `sed` commands from `process_term` returning exit-code 1, the `process_term` function still continues to completion. This stops when I remove the if-block around `process_term`.

> As more comprehensive alternative, we could introduce something like this:
> ```shell
> sed_msg() {
>  local msg=$1
>  shift
>
>  if ! sed "$@"; then
>    echo "Failed $msg" >&2
>    exit 1
>  fi
> }
> ```

Works like a charm. Modified the PR.

WDYT?

pbrisbin (approved):
Should we add a test case?

atreyasha:
Yes, done

PR #196, opened 2022-04-18
Integrating `pacignore`

atreyasha:
### Checklist

* Admin:
  - [x] No duplicate PRs
* Integration:
  - [x] Update unit tests
 

### Description

This PR aims to integrate `pacignore` into `downgrade` as discussed previously. So far I did the following:

1. Migrate `downgrade` and `pacignore` scripts into dedicated `bin` directory
2. Update `Makefile` to install both scripts given new locations
3. Update unit tests with new paths and specifications for `pacignore` as discussed previously

@pbrisbin WDYT about the diffs so far?

### Next steps

Besides the usual housekeeping tasks in the checklist above, I want to add the following features in this PR:

~1. Add an `includes` subcommand to `pacignore` to check if a package is currently being ignored, for example `pacignore includes foo` to check if `foo` is currently being ignored. This feature would simplify the `prompt_to_ignore` function in the `downgrade` script and would reduce some boilerplate code in the `pacignore` script.~
1. Add `pacignore` invocations inside the `prompt_to_ignore` function in the `downgrade` script
2. Update GitHub workflows and `restyled` to also check `pacignore`

Regarding implementing point 1 above, I plan to change the `prompt_to_ignore` function to the following:

```shell
prompt_to_ignore() {
  local pkg ans

  for pkg; do
    pacignore check --pacman-conf "$PACMAN_CONF" "$pkg" && return 0
    eval_gettext "add \$pkg to IgnorePkg? [y/N] "
    read -r ans
    if [[ "${ans,,}" == $(gettext 'y')* ]]; then
      pacignore add --pacman-conf "$PACMAN_CONF" "$pkg"
    fi
  done
}
```

I tested this locally and it works fine with a system-level install. However, I am not sure how to make this work successfully for the unit tests. This is because the current unit tests source functions from `downgrade` and `pacignore` and are not able to execute scripts _per-se_ (such as `pacignore` in the example above) because we export `LIB=1` in `helper*.sh`.

Does it make sense to mask `pacignore` with a no-op for the `test/prompt_to_ignore` unit tests, since we already have dedicated unit tests for `pacignore` in `test/pacignore`? That way, we would only be testing how `prompt_to_ignore` behaves with "y" or "n" answers and would keep everything related to `pacignore` separate.


bin/pacignore
```diff
@@ -0,0 +1,238 @@
...
+  # Proceed with main workflow
+  main "${pkgs[@]}"
+}
+
+# Set script defaults
+PACMAN_CONF="/etc/pacman.conf"
+PACIGNORE_SUBCOMMAND="ls"
+
+# CLI execution (not for testing)
+if ((!LIB)); then
```
pbrisbin, line 200: Part of the appeal extracting separate executables, is you now have a smaller unit you can test -- without using `source` and then calling internal functions. `pacignore` is pretty simple, and I think could be unit-tested directly without this machinery.

```
It lists the thing,

    $ echo "whatever" > something.conf
    > PACMAN_CONF=something.conf pacignore ls
    whatever
``` 

Know what I mean?

IMO, the fact that `downgrade` is tested the way it is is a bit of an anti-pattern, so I wouldn't want to spread it to another script like this.

bin/pacignore
```diff
@@ -0,0 +1,238 @@
...
+  if [[ -d "$locale" ]]; then
+    # Packaged installation
+    TEXTDOMAINDIR=$(cd "$locale" && pwd)
+  else
+    # Probably testing ./pacignore
+    TEXTDOMAINDIR=/usr/share/locale
+  fi
+
+  # Export detected environmental variables
+  export TEXTDOMAIN=pacignore TEXTDOMAINDIR
```
pbrisbin, line 217: So much boilerplate to localization. I wish we could minimize this, now that's being used from separate scripts; but I just don't know how.

bin/pacignore
```diff
@@ -0,0 +1,238 @@
...
+  # Check for help option
+  for arg; do
+    if [[ "$arg" =~ ^-h$|^--help$ ]]; then
+      usage
+      exit 0
+    fi
+  done
+
+  # Check if run as root
+  if ((UID)); then
```
pbrisbin, line 228: I don't know that we should explicitly forbid non-root for this. I would let file permissions errors raise themselves if they occur. An `ls` operation, for example, should work fine as non-root.

bin/pacignore
```diff
@@ -0,0 +1,238 @@
...
+
+parse_options() {
+  # Parse the subcommand correctly
+  if [[ "$1" =~ ^(ls|check|add|rm)$ ]]; then
+    PACIGNORE_SUBCOMMAND="$1"
+    shift
+  fi
+
+  # Loop across all CLI arguments and categorize
+  while [[ -n "$1" ]]; do
```
pbrisbin, line 142: I remember I suggested we use Bash's `getopt` instead of rolling our own. This was because, while I'm familiar with (and enjoy) POSIX's `getopts`, I know that it doesn't support long options, but the Bash one does.

I've done some reading since then, and I think it's a mistake. Specifically https://mywiki.wooledge.org/ComplexOptionParsing.

It seems Bash's `getopt` is a mess and no one should use. And the solution for supporting long options is to indeed roll your own. That said, I still prefer POSIX's `getopt` over that, and simply living with single-letter-options-only as a limitation.

The ship has sailed on `downgrade`, IMO, but perhaps as we break out new tools we could go single-letter with POSIX `getopts`?

pbrisbin (commented):
I left a few thoughts, but will hold further review for now.

One high-level thought I had is that we're extracting a new (smaller, simpler) executable here, but we're still doing a lot of the "noisy" things we do in `downgrade` that were specifically necessary because of its higher level of complexity.

Personally, I would seek to not have many functions, not have many options, not do the `LIB`/`main`/`cli` dance, etc, etc. My comment about testing from the outside is basically this point: we should be able to test this fully by supplying a fake `PACMAN_CONF`, calling `ls`-and-friends, and checking that file's content. If tested that way, we don't need a ton of the structural complexity of the functions and `$LIB`-etc.

WDYT?

bin/pacignore
atreyasha: Hmm, I see and did not know about limitations regarding Bash's `getopt` either. I am not against going with POSIX's `getopts` with single-letter options, but I also think we should work on this migration in a separate PR (addressing #102) for all our source code sitting in the `bin` directory.

Would this be fine?

pbrisbin: >  I also think we should work on this migration in a separate PR

I was not suggesting a migration of `downgrade`s options handling (now or ever), I was suggesting that as we make new things (such as this PR) we make them with that kind of options-handling from the start.

Are you saying you'd prefer to even land this PR as-is and then migrate it later?

bin/pacignore
atreyasha: Makes sense. Fixed this now

bin/pacignore
atreyasha: Got it and changed this. Now `pacignore` is tested on a script-level

bin/pacignore
atreyasha: I can't see any way of getting around this part if we want `pacignore` to function as a standalone script and still offer localization support

bin/pacignore
atreyasha: > I was suggesting that as we make new things (such as this PR) we make them with that kind of options-handling from the start

Ah I see. Hmm, then the question would be whether we start writing the new options-handling now or do we start doing it when we split `downgrade` into the suite of smaller scripts i.e. `downgrade-select` and family.

I see the benefit of doing it now in that we already start this process. I also a downside of this PR becoming too large IMO.

> Are you saying you'd prefer to even land this PR as-is and then migrate it later?

Yes, I would prefer to finish this PR as-is and work on the migration in a separate PR; preferably in the same PR or step when we start splitting `downgrade` into its subordinate scripts. WDYT?

atreyasha:
Apologies for the late response.

> One high-level thought I had is that we're extracting a new (smaller, simpler) executable here, but we're still doing a lot of the "noisy" things we do in downgrade that were specifically necessary because of its higher level of complexity.
> WDYT?

That's true, I initially just modeled the syntax and flow in `pacignore` from that of `downgrade` and did not put too much thought into simplicity. But I agree, ultimately we are working towards making things "simpler" and what you mentioned is indeed a step towards that.

Let me know your thoughts on the changes made in this direction

atreyasha:
> Does it make sense to mask pacignore with a no-op for the test/prompt_to_ignore unit tests, since we already have dedicated unit tests for pacignore in test/pacignore? That way, we would only be testing how prompt_to_ignore behaves with "y" or "n" answers and would keep everything related to pacignore separate.

Going back to what I mentioned in the PR description, I went ahead and made this change in the latest commit. Now `pacignore` is called by `downgrade` and I update the unit tests of `downgrade` to reflect this as well.

bin/pacignore
atreyasha: > I was not suggesting a migration of downgrades options handling (now or ever)

To this point, I would prefer our CLI option-parsing method to be consistent among our scripts. By this I mean that if we use POSIX `getopts` in `pacignore`, then I think we should be using it in all of our scripts. This is another reason why I am for doing the migration outside of this PR, since it would (in this assumption) involve migration of multiple scripts

pbrisbin: > I would prefer our CLI option-parsing method to be consistent among our scripts

Aha, that's where we disagree, I think. I don't see any relation between the scripts -- they're just things that call each other, `downgrade` calls `pacignore` no differently than `git` might call `cat` or something like that. Those two don't need any "consistency", right?

This is less true of the way `downgrade` will make use of `downgrade-x` scripts; they have a clear relationship. However, in that case I would look at it as a "high-level" (`downgrade`) vs "low-level" (`downgrade-x`) interface; and I think it'd be reasonable for the high-level to use "nice" (i.e. long flags) options parsing, while the low-level might use something more spartan but easier to implement. So that's one reason I'd be OK with inconsistency between these two.

>  I also see a downside of this PR becoming too large IMO.

But I'm suggesting you do the options parsing only in `pacignore` a simpler way. It would make this PR _smaller_, no? If you can't tell, I'm big on compartmentalization and isolation to reduce cognitive load.

I'm reviewing this PR as establishing a distinct `pacignore` script and not concerning myself with any changes to `downgrade`. In fact, if you were interested in smaller PRs, you could remove everything else and do all the packaging changes and updates to `downgrade` to _use_ `pacignore` in another PR after :thinking: 

All that said, I don't feel strongly on these points and am happy to Approve the PR to let the work evolve. I just wanted to clarify my position to make sure we were on the same page.

pbrisbin (approved):

bin/pacignore
atreyasha: > So that's one reason I'd be OK with inconsistency between these two.

Thanks for explaining with that reference. I meant consistency in the sense of having a common option parsing standard across all our scripts. But if you frame `downgrade` as a "higher-level" script calling other "lower-level" scripts, then yes I agree it makes sense to have a simpler option parsing framework for the "lower-level" scripts.

Simpler framework aside, another thing holding me back from POSIX's `getopts` was that I have never used it before. Could you perhaps point me to some good starting documentation or suggest a basic diff that could have `getopts` working in `pacignore`?

> It would make this PR smaller, no? 

My bad with a wrong word choice. By "smaller" I meant to minimize the effort in this PR so as to keep things moving and maintain a certain momentum. As mentioned before, I have never used `getopts` so using it in this PR would increase the effort for me to learn and implement it. But this is not an issue since I am on-board for using `getopts`; I just wanted to explain my thought process better.

> If you can't tell, I'm big on compartmentalization and isolation to reduce cognitive load.

Yes

> In fact, if you were interested in smaller PRs, you could remove everything else and do all the packaging changes and updates to downgrade to use pacignore in another PR after thinking

My usual PR workflow is to have enough packed in the PR to allow for a meaningful release afterwards (regardless of whether we actually release or not). This is also why I added the PR checklist with points pertaining to releases. But if you find that this format is too convoluted and increases "cognitive workload", I would be glad to discuss and find an alternative such as having certain PRs for core code changes and other PRs for housekeeping tasks.

PR #197, opened 2022-05-15
Integrate `pacignore` into `downgrade`

atreyasha:
### Checklist

* Admin:
  - [x] No duplicate PRs
* Integration:
  - [x] Update unit tests

### Description

To work towards smaller and more compact PRs, I moved (and split) a commit from the `as/pacignore` branch to this branch. This PR proposes the following:

1. Use `pacignore` directly in `downgrade` as a form of integration
2. Simplify unit tests of `downgrade` to no longer test features of what is now `pacignore`. Instead, we only test integration between the two scripts.

### Pending

- [x] Use short CLI options for `pacignore` in this PR once #198 is merged

### DIFF
--- a/bin/downgrade
+++ b/bin/downgrade
@@ -110,15 +110,23 @@ read_selection() {
   ((ans > 0 && ans <= $#)) && printf "%s" "${!ans}"
 }
 
+prompt_gettext() {
+  local key="$1" ans
+
+  eval_gettext "$key"
+  read -r ans
+
+  [[ "${ans,,}" == $(gettext 'y')* ]]
+}
+
 prompt_to_ignore() {
-  local pkg ans
+  local pkg
 
   for pkg; do
-    pacignore check --pacman-conf "$PACMAN_CONF" "$pkg" && return 0
-    eval_gettext "add \$pkg to IgnorePkg? [y/N] "
-    read -r ans
-    if [[ "${ans,,}" == $(gettext 'y')* ]]; then
-      pacignore add --pacman-conf "$PACMAN_CONF" "$pkg"
+    if ! pacignore check -c "$PACMAN_CONF" "$pkg"; then
+      if prompt_gettext "add \$pkg to IgnorePkg? [y/N] "; then
+        pacignore add -c "$PACMAN_CONF" "$pkg"
+      fi
     fi
   done
 }
--- a/bin/downgrade
+++ b/bin/downgrade
@@ -111,34 +111,14 @@ read_selection() {
 }
 
 prompt_to_ignore() {
-  local pkg ans ln
+  local pkg ans
 
   for pkg; do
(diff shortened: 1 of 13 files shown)


bin/downgrade
```diff
@@ -111,34 +111,14 @@ read_selection() {
 }
 
 prompt_to_ignore() {
-  local pkg ans ln
+  local pkg ans
 
   for pkg; do
-    grep -Eq '^IgnorePkg.*( |=)'"$pkg"'( |$)' "$PACMAN_CONF" && return 0
-
+    pacignore check --pacman-conf "$PACMAN_CONF" "$pkg" && return 0
```
atreyasha, line 117: I see a potential failing case here because of `return 0`. For example, if we execute `downgrade foo bar` and `foo` is already being ignored; then the prompt for ignoring `bar` will never be executed since `prompt_to_ignore` will return 0 upon checking `foo`.

The change below should fix this and I should also add a test case for this.

```suggestion
    pacignore check --pacman-conf "$PACMAN_CONF" "$pkg" && continue
```

atreyasha: Thanks, changed it

atreyasha:
Wanted to get your advice on something.

ATM, I am masking the `pacignore` call from `downgrade` during testing to keep the latter's unit tests simple:

https://github.com/archlinux-downgrade/downgrade/blob/76b894cf9481376e867f21f99f75475aa681e86b/test/helper.sh#L31

Ideally, I would like to not do this and instead test the full `pacignore` call from `downgrade`. However, this would require a more complicated diff to the `downgrade` tests such as updating the path to include `pacignore` and overriding the `write_pacman_conf` function for this specific test case with something like this below:

https://github.com/archlinux-downgrade/downgrade/blob/76b894cf9481376e867f21f99f75475aa681e86b/test/helper_pacignore.sh#L8-L12

I see two possible options going forward:

1. Implement this diff now to have a proper integration test between `downgrade` and `pacignore`
2. Wait until we split `downgrade` into lower-level scripts and then implement integration tests for all lower-level scripts called from `downgrade`

An advantage of (1) is that we have the integration test done to check for bugs during development. A disadvantage of (1) is that it might involve work that we would discard later with the `downgrade` split, since we would anyways be moving most tests from function-level to script-level (as per my understanding).

WDYT?

atreyasha:
Thanks for explaining this in detail.

> These unit tests with mocking would be my ideal, but it's not easy in a shell project.

Yes, that sounds like the best case scenario but I also see what you mean about the additional complexity involved.

> but I'm happy to defer to you and/or continue to help you explore these alternatives for yourself

I'll think a bit more about this and see if there could be an easy + maintainable way of having unit tests for `downgrade` + mocking for lower-level scripts. Otherwise, we can still work with what we have now.

Thanks again, wishing you a nice Friday! :)

PR #200, opened 2022-05-22
Pacignore housekeeping

atreyasha:
### Checklist

* Admin:
  - [x] No duplicate PRs
* Integration:
  - [ ] Update shell completions
* Documentation:
  - [ ] Update mandoc
  - [ ] Update locales
* Release:
  - [ ] Update change log

### Description

This PR aims at completing housekeeping tasks which are shown above as check-boxes. Currently, I have updated and simplified the `Makefile` with pattern rules and shell for-loops (plus tested installation and uninstallation locally). I'll work on the other tasks gradually.

For now, WDYT about the changes to the `Makefile`?


Makefile
```diff
@@ -3,59 +3,63 @@ LOCALEPREFIX ?= $(PREFIX)/share/locale
...
-locale/downgrade.pot: bin/downgrade
+locale/%.pot: bin/%
 	xgettext \
 		--from-code=utf-8 -L shell \
-		--package-name=downgrade \
+		--package-name=$* \
 		--copyright-holder=$(AUTHOR) \
 		-o $@ $<
-	find ./locale/*po -exec msgmerge --update {} $@ \;
+	find ./locale/$*/*.po -exec msgmerge --update {} $@ \;
```
pbrisbin, line 20: Could use `-name '*.po'`, which is a touch clearer IMO -- but functions the same.

atreyasha: True, modified

Makefile
```diff
@@ -3,59 +3,63 @@ LOCALEPREFIX ?= $(PREFIX)/share/locale
...
 
 .PHONY: locale
-locale: locale/downgrade.pot
+locale: $(SCRIPTS:%=locale/%.pot)
 
-doc/downgrade.8: doc/downgrade.8.md
+doc/%.8: doc/%.8.md
```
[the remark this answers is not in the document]
atreyasha: Ahh yes good catch!

Makefile
```diff
@@ -3,59 +3,63 @@ LOCALEPREFIX ?= $(PREFIX)/share/locale
...
 	$(PANDOC) --standalone --to man $< -o $@
 
 .PHONY: man
-man: doc/downgrade.8
+man: $(SCRIPTS:%=doc/%.8)
```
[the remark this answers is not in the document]
atreyasha: Makes sense, I did it in a dynamic way

Makefile
```diff
@@ -3,59 +3,63 @@ LOCALEPREFIX ?= $(PREFIX)/share/locale
...
-	install -Dm755 bin/pacignore $(DESTDIR)$(PREFIX)/bin/pacignore
 	install -Dm644 conf/downgrade.conf $(DESTDIR)$(XDGPREFIX)/downgrade/downgrade.conf
-	install -Dm644 doc/downgrade.8 $(DESTDIR)$(MANPREFIX)/man8/downgrade.8
-	install -Dm644 completion/bash $(DESTDIR)$(PREFIX)/share/bash-completion/completions/downgrade
-	install -Dm644 completion/zsh $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_downgrade
-	install -Dm644 completion/fish $(DESTDIR)$(PREFIX)/share/fish/vendor_completions.d/downgrade.fish
-	for po_file in locale/*.po; do \
+	for script in $(SCRIPTS); do \
+	  install -Dm755 bin/$$script $(DESTDIR)$(PREFIX)/bin/$$script; \
+	  install -Dm644 doc/$${script}.8 $(DESTDIR)$(MANPREFIX)/man8/$${script}.8; \
```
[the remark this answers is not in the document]
atreyasha: Makes sense. I pulled the man pages out of the loop, thanks

Makefile
```diff
@@ -3,59 +3,66 @@ LOCALEPREFIX ?= $(PREFIX)/share/locale
...
-	install -Dm644 completion/zsh $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_downgrade
-	install -Dm644 completion/fish $(DESTDIR)$(PREFIX)/share/fish/vendor_completions.d/downgrade.fish
-	for po_file in locale/*.po; do \
+	install -Dm644 doc/pacignore.8 $(DESTDIR)$(MANPREFIX)/man8/pacignore.8
+	for script in $(SCRIPTS); do \
+	  install -Dm755 bin/$$script $(DESTDIR)$(PREFIX)/bin/$$script; \
+	  install -Dm644 completion/$$script/bash $(DESTDIR)$(PREFIX)/share/bash-completion/completions/$$script; \
+	  install -Dm644 completion/$$script/zsh $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_$${script}; \
+	  install -Dm644 completion/$$script/fish $(DESTDIR)$(PREFIX)/share/fish/vendor_completions.d/$${script}.fish; \
+	done
```
atreyasha, line 46: Two questions on (un)installation targets:

1. Do you think this for-loop over `$(SCRIPTS)` is fine or should we also write this out line-by-line (as we did for the mandocs)?
2. Should we quote all paths in `Makefile` targets?

Makefile
```diff
@@ -3,59 +3,66 @@ LOCALEPREFIX ?= $(PREFIX)/share/locale
...
 	install -Dm644 conf/downgrade.conf $(DESTDIR)$(XDGPREFIX)/downgrade/downgrade.conf
 	install -Dm644 doc/downgrade.8 $(DESTDIR)$(MANPREFIX)/man8/downgrade.8
-	install -Dm644 completion/bash $(DESTDIR)$(PREFIX)/share/bash-completion/completions/downgrade
-	install -Dm644 completion/zsh $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_downgrade
-	install -Dm644 completion/fish $(DESTDIR)$(PREFIX)/share/fish/vendor_completions.d/downgrade.fish
-	for po_file in locale/*.po; do \
+	install -Dm644 doc/pacignore.8 $(DESTDIR)$(MANPREFIX)/man8/pacignore.8
+	for script in $(SCRIPTS); do \
+	  install -Dm755 bin/$$script $(DESTDIR)$(PREFIX)/bin/$$script; \
+	  install -Dm644 completion/$$script/bash $(DESTDIR)$(PREFIX)/share/bash-completion/completions/$$script; \
```
atreyasha, line 43: Two questions regarding (un)installation targets:

1. Were you suggesting to replace the for-loop over `$(SCRIPTS)` with individual lines too? I don't see a major disadvantage in the for-loop here because we intend to install all executables in `bin` and each executable should have shell completions and locales associated with it. But maybe there are edge cases and then individual lines make more sense.
2. Should we quote all paths in `Makefile` targets?

pbrisbin: > Were you suggesting to replace the for-loop for installing executables

I wasn't, I'm fine either way on that one. The same arguments as man pages do apply: it's a small rarely-changing list, so I would probably write it out myself. However, it was dealing with the `man8` situation that made me suggest removing that loop, which doesn't apply to the executables one.

> Should we quote

I doesn't hurt, but I don't think we need to. We fully control all of the files involved here and know they will never contain spaces.

pbrisbin (approved):

Makefile
atreyasha: > I wasn't, I'm fine either way on that one

Perfect, then will keep that loop

> I doesn't hurt

Quoted the paths and arguments which have shell/Makefile variables just to be on the safe side, though I agree this should never happen in normal cases

atreyasha (commented):
Added shell completions for `pacignore` and tested them all locally as well. Let me know if you have suggestions for improvements.

doc/pacignore.8.md
```diff
@@ -0,0 +1,66 @@
...
+
+> Remove package(s) from the IgnorePkg directive.
+
+# OPTIONS
+
+**\-c** _\<path\>_\
+
+> Pacman configuration file, default is _/etc/pacman.conf_.
+
+**\-h**\ _\<flag\>_\
```
atreyasha, line 36: Added a mandoc for `pacignore`.

Question regarding syntax:

CLI options such as `-h` don't require an argument, so they should not require an argument type enclosed by the `<*>` characters. However, `-h` renders quite strange in the mandoc without an argument due to lots of space between it and the description below. So I added a "phantom" argument called `<flag>` to indicate this is just a flag and nothing more.

Is it fine if we do it this way (also for other similar cases)?

pbrisbin: > So I added a "phantom" argument ... this is just a flag

Hmm. I think it implies `-h` "takes a flag", whatever that might mean. So it feels confusing to me.

> -h renders quite strange in the mandoc without an argument due to lots of space between it and the description below

I did a small test in another project and I see,



This seems totally fine to me. Sure there's some space, but it's all consistent.

That said, you could certainly do something like,

```markdown
**\-h** Show help message
```

And get



:shrug:

atreyasha:
As this PR is getting quite large (both in terms of effort and diff size) and it is only going to get larger with changes to locales, I will merge it now and open a new PR for the next steps

PR #202, opened 2022-05-27
Pacignore housekeeping (part 2)

atreyasha:
### Checklist

* Admin:
  - [x] No duplicate PRs
* Documentation:
  - [ ] Update locales
* Release:
  - [ ] Update change log

### Description

In this PR, I aim finish up all tasks related to locales and if possible release with a minor version bump.

### DIFF
--- a/bin/downgrade
+++ b/bin/downgrade
@@ -110,21 +110,14 @@ read_selection() {
   ((ans > 0 && ans <= $#)) && printf "%s" "${!ans}"
 }
 
-prompt_gettext() {
-  local key="$1" ans
-
-  eval_gettext "$key"
-  read -r ans
-
-  [[ "${ans,,}" == $(gettext 'y')* ]]
-}
-
 prompt_to_ignore() {
-  local pkg
+  local pkg ans
 
   for pkg; do
     if ! pacignore check -c "$PACMAN_CONF" "$pkg"; then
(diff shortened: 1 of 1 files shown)


bin/downgrade
```diff
@@ -110,21 +110,14 @@ read_selection() {
   ((ans > 0 && ans <= $#)) && printf "%s" "${!ans}"
 }
 
-prompt_gettext() {
-  local key="$1" ans
-
-  eval_gettext "$key"
```
atreyasha, line 116: As a quick update, I had to revert this simplification to `downgrade`'s call of `pacignore` because `xgettext` was not able to resolve this line and would assume that this prompt no longer requires localization support. IMO this is not a big loss because the reverted portion still looks compact and reads well IMO

Makefile
```diff
@@ -61,7 +61,7 @@ uninstall:
 	    "$(DESTDIR)$(PREFIX)/share/bash-completion/completions/$$script" \
 	    "$(DESTDIR)$(PREFIX)/share/zsh/site-functions/_$${script}" \
 	    "$(DESTDIR)$(PREFIX)/share/fish/vendor_completions.d/$${script}.fish" \
-	    "$(DESTDIR)$(LOCALEPREFIX)/*/LC_MESSAGES/$${script}.mo"; \
```
atreyasha, line 64: Classic, forgot to unquote the asterisk

pbrisbin (approved):

PR #215, opened 2023-03-31
Patch ZSH completions

atreyasha:
### Checklist

* Admin:
  - [x] No duplicate PRs
* Integration:
 

### Description

Following #211, I thought it would be a good idea to implement ZSH completion caching for `pacignore` (as we have it now for `downgrade`). This PR does the following:

1. 88fa098887f04514ede7d979f56f306cdcc36d19: Implements the ZSH completion caching feature for `pacignore`
2. f13c35772fcf3dee208394ffb314a51197fa7135: Potentially fixes a bug in the caching for both `downgrade` and `pacignore` by checking if `pacman.log` is newer than the cache, instead of vice versa that was done before

@pbrisbin could you help me check if the fix in point (2) makes sense? I could not find much documentation on the ZSH completion caching policy...

### DIFF
--- a/completion/downgrade/zsh
+++ b/completion/downgrade/zsh
@@ -1,15 +1,15 @@
 #compdef downgrade
 
 _downgrade_caching_policy () {
-  [[ "/var/log/pacman.log" -nt "$1" ]]
+  [[ '/var/log/pacman.log' -nt "$1" ]]
 }
 
 _set_downgrade_cache_policy () {
   local update_policy
-  zstyle -s ":completion:*:*:downgrade:*" cache-policy update_policy
+  zstyle -s ':completion:*:*:downgrade:*' cache-policy update_policy
 
   if [[ -z "$update_policy" ]]; then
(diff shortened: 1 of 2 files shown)


pbrisbin (approved):
LGTM.

I, unfortunately, I have no idea about your question. This is all black magic to me!

atreyasha:
Thanks! Documentation was quite obscure but found some justification that this fix works:

`zstyle` syntax: https://zsh.sourceforge.io/Doc/Release/Zsh-Modules.html#The-zsh_002fzutil-Module and https://thevaluable.dev/zsh-completion-guide-examples/

Examples from other packages:

`bazel`: https://bazel.googlesource.com/bazel/+/master/scripts/zsh_completion/_bazel
`gradle`: https://github.com/aosm/zsh/blob/master/zsh/Completion/Unix/Command/_gradle
`zulu`: https://github.com/zulu-zsh/zulu/blob/master/zulu.zsh-completion

PR #217, opened 2023-10-04
automatic package selection

krumelmonster:
### Checklist

* Admin:
  - [x] No duplicate PRs
* Integration:
  - [x] Update unit tests
  - [ ] Update configuration file
* Documentation:
  - [x] Update usage function
  - [x] Update mandoc
  - [x] Update readme
  - [x] Update locales
* Release:
  - [ ] Update change log

### Description

The `--any-candidate` option picks the first candidate instead of presenting options with fzf. In conjunction with `--ignore never` (or always) it allows for noninteractive downgrades. It may not be a complete fix for https://github.com/archlinux-downgrade/downgrade/issues/118 but it's enough to set a bunch of packages to a version of choice and not care whether they are sourced locally or from the archives.


bin/downgrade
```diff
@@ -290,6 +291,8 @@ process_term() {
     } >&2
   elif (("${#candidates[@]}" == 1)); then
     choice=${candidates[0]}
+  elif ((DOWNGRADE_TO_ANY)); then
+    choice=${candidates[0]}
```
pbrisbin, line 295: I worry about the word "any" and description "just pick any candidate", since that doesn't indicate how the choice is made. It's like "oh, just pick _any_ candidate, I don't care which". I forget which way the results are ordered, but whichever `[0]` is (newest or oldest), I would indicate that in the option name and description somehow.

Bonus points for supporting both via (e.g.) `--select=oldest|newest`:

```bash
elif [[ -n "$DOWNGRADE_SELECT" ]]; then
  case "$DOWNGRADE_SELECT" in
    oldest) choice=${candidates[0]} ;;
    newest) choice=${candidates[-1]} ;;
    *) # error
    ;;
  esac
```

krumelmonster: Good point. I have made sure it will pick a version from the cache if there is one. Right now it does pick the oldest which is probably not what people want. In my use case I did specify precise versions so I didn't care initially:

https://github.com/archlinux-downgrade/downgrade/issues/118#issuecomment-1746830312

I will change it to picking the newest matching version for now but then this would still be the newest locally available version (if applicable) and the newest ala version otherwise

krumelmonster: The readme now documents the behaviour like this

> Never prompt for version or location, pick automatically instead. Allows for
> noninteractive downgrades when used in conjunction with --ignore
> If a package matching the version filter is found in the cache, pick the
> latest matching version from cache. Íf no matches are found in cache or
> --ala-only was specified, then pick the latest matching version from ala.

While `--help` and completion just call it "automatic package selection"

We can rename the option to `--autoselect`. I would refrain from --select=oldest|newest because if the version is not specifically mentioned, newer is better and I think the behaviour of picking from the cache before ala makes a lot of sense for when both are enabled.

README.md
```diff
@@ -37,6 +37,9 @@ Options:
   --cached-only   only use cached packages
   --ignore        <prompt|always|never>
                   whether to add packages to IgnorePkg
+  --latest        pick latest matching version
+  --oldest        pick oldest matching version
+  --cache-preferred do not query ala if a matching package was found in cache
```
pbrisbin, line 42: I see `--prefer-cache` as the naming elsewhere, I think I like that better.

bin/downgrade
```diff
@@ -492,6 +523,9 @@ DOWNGRADE_MAXDEPTH=1
 DOWNGRADE_CONF="/etc/xdg/downgrade/downgrade.conf"
 DOWNGRADE_VERSION="11.3.0"
 DOWNGRADE_IGNORE="prompt"
+DOWNGRADE_TO_LATEST=0
+DOWNGRADE_TO_OLDEST=0
```
pbrisbin, line 527: I wonder if we should do `DOWNGRADE_SELECT=latest|oldest`, since `DOWNGRADE_LATEST=1 DOWNGRADE_OLDEST=1` is an invalid state. We could still use `--latest` and `--oldest` CLI options, but at least in terms of ENV have something a bit more robust. WDYT?

doc/downgrade.8.md
```diff
@@ -91,6 +91,17 @@ on the ALA.
...
+**\--latest | --oldest**\
+
+> Never prompt for version or location, automatically pick the most up to
+> date/most out of date version. Allows for noninteractive downgrades when
+> used in conjunction with --ignore
+
+**\--cache-preferred**\
+
+> If a package matching the version filter is found in the cache, skip querying
+> ala.
```
pbrisbin, line 103: I would mention the relationship to `--cached-only`, adding something like "this is like `--cached-only` accept it will still try ALA if no package is found."

bin/downgrade
krumelmonster: They are not really env, they are just internal variables. I'll add some logic to make --latest and --oldest mutually exclusive -- this I think will be easier when using separate DOWNGRADE_TO_LATEST/DOWNGRADE_TO_OLDEST variables so I'd stick with them

krumelmonster: [7179aac](https://github.com/archlinux-downgrade/downgrade/pull/217/commits/7179aace86e6c841992052c6e450cfa6c8edbd82) makes it consistent with how DOWNGRADE_FROM_ALA and DOWNGRADE_FROM_CACHE work as "mutually exclusive".

README.md
```diff
@@ -37,6 +37,9 @@ Options:
   --cached-only   only use cached packages
   --ignore        <prompt|always|never>
                   whether to add packages to IgnorePkg
+  --latest        pick latest matching version
+  --oldest        pick oldest matching version
+  --prefer-cache do not query ala if a matching package was found in cache
```
pbrisbin, line 42:
```suggestion
  --prefer-cache  do not query ala if a matching package was found in cache
```

bin/downgrade
```diff
@@ -27,6 +27,9 @@ $(gettext "Options"):
   --cached-only   $(gettext "only use cached packages")
   --ignore        <prompt|always|never>
                   $(gettext "whether to add packages to IgnorePkg")
+  --latest        $(gettext "pick latest matching version")
+  --oldest        $(gettext "pick oldest matching version")
+  --prefer-cache $(gettext "do not query ala if a matching package was found in cache")
```
pbrisbin, line 32:
```suggestion
  --prefer-cache  $(gettext "do not query ala if a matching package was found in cache")
```

bin/downgrade
```diff
@@ -281,7 +286,18 @@ process_term() {
...
   installed=($(previously_installed "$name"))
   current=$(currently_installed "$name")
-  candidates=($(search_packages "$name" | filter_packages "$name" "$operator" "$version" | sort_packages))
+
+  candidates=()
+  if ((DOWNGRADE_FROM_CACHE)); then
+    candidates+=($(search_cache "$name" | filter_packages "$name" "$operator" "$version"))
+  fi
+  if ((DOWNGRADE_FROM_ALA)); then
+    if ((DOWNGRADE_PREFER_CACHE == 0)) || (("${#candidates[@]}" == 0)); then
```
pbrisbin, line 295: I believe this is the equivalent combined condition,

```suggestion
  if ((DOWNGRADE_FROM_ALA)) && { ((!DOWNGRADE_PREFER_CACHE)) || ((!"${#candidates[@]}")) }; then
```

pbrisbin (approved):

krumelmonster:
Can we merge or should I fixup these minor fixes first?

doc/downgrade.8.md
```diff
@@ -91,6 +91,18 @@ on the ALA.
...
+
+> Never prompt for version or location, automatically pick the most up to
+> date/most out of date version. Allows for noninteractive downgrades when
+> used in conjunction with `--ignore`.
+
+**\--prefer-cache**\
+
+> If a package matching the version filter is found in the cache, skip querying
+> ala. Unlike --cached-only, --prefer-cache does query ala if no package could
+> be matched in cache.
```
pbrisbin, line 104: You need to escape **all** the `--`s in here. Otherwise, the man-page renders wrong. You can see the `\[en]`s above, which we want to avoid.

pbrisbin (approved):
This LGTM except for fixing up the escaped `--` in the man-page. I'll merge once that's fixed.

pbrisbin:
> I'll merge once that's fixed.

Oops, and the test failures of course. Seems there's a syntax error somewhere :thinking:

pbrisbin:
Let me know when you've gotten tests passing, or if you need any help with that.

krumelmonster:
whoops this may have been a bit too much force pushing. I only made the minor changes we've mentioned before but then I also rebased a bit.
https://github.com/archlinux-downgrade/downgrade/compare/d7e2077597c8429c5ec2f9d319da5170dab7549b..b80204251a70b26e2a77eeb1a3008829ce18ed51

krumelmonster:
I updated the tests that broke due to the split-up of search_packages into search_ala and search_cache and rebased into the respective commit https://github.com/archlinux-downgrade/downgrade/pull/217/commits/d82aa216c1698f92e8409552fea0eb84910bb6d6 and I think we're now good to go. Workflows should succeed once you run them; tested here https://github.com/krumelmonster/downgrade/pull/1

I'm tempted to add some test cases for my changes themselves (cram looks really cool!) but that will be a task for another day (preferably in a later pr).

Again Sorry for the rebase mess and the mangling with https://github.com/archlinux-downgrade/downgrade/pull/231 , changes since I started rebasing are:
 - test cases
 - syntax error due to missing ; in the {} (interestingly only CI catched this and bash ate it on my local arch)
 - escaping the \-- in the manpage

overall diff since before first force push: https://github.com/archlinux-downgrade/downgrade/compare/d7e2077597c8429c5ec2f9d319da5170dab7549b..d75b63242be75441c9891e659f52cff9b14e71ba

PR #248, opened 2025-04-02
fix(release): automate PKGBUILD maintenance

pbrisbin:
Extracted `archlinux-downgrade/aur-publish-action` to handle things.


pbrisbin:
This PR looks good. Just going to polish up the action, auto-release that, then one final test here.

PR #254, opened 2025-06-06
fix: avoid empty-string argument when config is empty

FlorianSchwarzl:
### Checklist

* Admin:
  - [x] No duplicate PRs

### Description

In newer versions of Bash (e.g., 5.2+), the following construct:

```bash
some_array=( $printf '%q ' $raw)
```

produces an array with a single empty string element (`("")`) when $raw is empty, whereas older versions (e.g., Bash 5.1 and earlier) would produce an empty array (`()`).

I'm not sure if this is actually Bash's fault, but the problem is real and the fix works (at least from my testing)

### DIFF
--- a/src/downgrade
+++ b/src/downgrade
@@ -56,8 +56,16 @@ read_pacman_conf() {
 
 read_downgrade_conf() {
   local var=$1
+  local raw
 
-  eval "$var=($(grep -E -v '^ *(#.*)?$' "$DOWNGRADE_CONF" 2>/dev/null | xargs printf '%q '))"
+  raw=$(grep -E -v '^ *(#.*)?$' "$DOWNGRADE_CONF" 2>/dev/null)
+
+  if [[ -z "$raw" ]]; then
+    # no real lines
+    eval "$var=()"
+  else
+    eval "$var=( $(printf '%q ' $raw) )"
+  fi
 }
 
 read_unique() {


pbrisbin (approved):

pbrisbin:
Looking into this a bit more, I'm having trouble reproducing.

We actually have a test for exactly this, and that test is passing. And when I run the equivalent operation directly, it also works as expected:

```console
[patrick@pele downgrade]$ bash --version | head -1
GNU bash, version 5.2.37(1)-release (x86_64-pc-linux-gnu)
[patrick@pele downgrade]$ echo "var=( $(grep -E -v '^ *(#.*)?$' test/conf/downgrade_empty.conf | xargs printf '%q '))"
var=(  )

```

`printf` does emit an empty element when given no arguments,

```console
[patrick@pele downgrade]$ printf '%q '
''
```

But the `grep` on an empty configuration file emits no lines, so I don't think the `xargs` or `printf` runs at all,

```console
[patrick@pele downgrade]$ grep -E -v '^ *(#.*)?$' test/conf/downgrade_empty.conf
[patrick@pele downgrade]$ grep -E -v '^ *(#.*)?$' test/conf/downgrade_empty.conf | xargs printf '%q '
 
```

Can you provide me the content your downgrade conf that hits this issue?

PR #264, opened 2026-02-14
Feat aur automatic fallback

ZorudaRinku:
### Checklist

* Admin:
  - [X] No duplicate PRs
* Integration:
  - [X] Always add unit tests for fixes or new functionality
* Documentation:
  - [X] If adding new options, update usage and `doc/*.ronn`
  - [X] If adding new strings, update `locale/*.po`

### Description

Add new feature; automatically falling back to AUR git downgrading when package is foreign to pacman.
Equivalent to manually passing `--git --pacman-cache ~/.cache/yay --maxdepth 2` if `pacman -Qm` returns 0 on a package.
Automatically checks for yay and paru in their default dirs to find git repos for downgrading.

### DIFF
--- a/src/downgrade
+++ b/src/downgrade
@@ -19,9 +19,8 @@ $(gettext "Options"):
                   $(gettext "pacman log file,")
                   $(gettext "default value taken from pacman configuration file,")
                   $(gettext "or otherwise defaults to") "/var/log/pacman.log"
-  --git           $(gettext "uses git for choosing version, viable only for aur packages")
   --maxdepth      <$(gettext "integer")>
-                  $(gettext "maximum depth to search for cached packages, defaults to") 1
+                  $(gettext "maximum depth to search for cached packages, defaults to") 2
   --ala-url       <url>
                   $(gettext "location of ALA server, defaults to") "https://archive.archlinux.org"
   --ala-only      $(gettext "only use ALA server")
@@ -37,6 +36,7 @@ $(gettext "Options"):
 
 $(gettext "Note"):
   $(gettext "Options after the -- characters will be treated as pacman options.")
+  $(gettext "yay and paru cache directories are searched automatically if present.")
   $(gettext "See downgrade(8) for details.")
 EOF
 }
@@ -80,6 +80,24 @@ previously_installed() {
     /.*\(installed\|upgraded\) \('"$1"'\) (\(.* -> \)\?\([^)]*\))/!d
     s//\2-\4/
   ' "$PACMAN_LOG"
+}
+
+get_real_user_home() {
+  local user=${SUDO_USER:-$USER}
+
+  if [[ -n "$user" ]]; then
+    getent passwd "$user" | cut -d: -f6
+  fi
+}
+
+detect_aur_caches() {
+  local home
+  home=$(get_real_user_home)
+
+  [[ -z "$home" ]] && return 0
+
+  [[ -d "$home/.cache/yay" ]] && printf '%s\n' "$home/.cache/yay"
+  [[ -d "$home/.cache/paru/clone" ]] && printf '%s\n' "$home/.cache/paru/clone"
 }
 
 currently_installed() {
@@ -168,8 +186,8 @@ matches_name_version_filter() {
     return 0
   fi
 
-  # version: 2025.11.12.r15.g0eed3fe
-  version_regex="[a-zA-Z0-9._]+"
+  # version: 1:2025.11.12.r15.g0eed3fe
+  version_regex="[a-zA-Z0-9._:]+"
   hash_or_minor="([0-9]+|[a-z0-9]{6,})"
   pkg_version=$(sed -r "s/.*$name-($version_regex)(-$hash_or_minor)?-(any|$DOWNGRADE_ARCH)\\.(git)?pkg\\.tar\\.(gz|xz|zst)/\1\2/g" <<<"$pkg")
   cmp=$(vercmp "$pkg_version" "$search_version")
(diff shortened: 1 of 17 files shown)


pbrisbin (changes requested):
This looks good but I'm wondering if we actually need any of these conditions / guard rails:

1. Package is foreign
2. AUR cache directory exists
3. AUR cache directory is not already given in `PACMAN_CACHE`

What happens if we check _none_ of that, and simply append the conventional AUR cache directories to `PACMAN_CACHE` as its default value at the start of the script, then let it just run as it does?

We may need to make `search_cache` tolerant of missing or duplicate cache directories, but I think that would be a better use of complexity.

ZorudaRinku:
This approach is so much simpler lol

- Had to update default maximum depth to 2, not sure if this is a huge problem but I don't think pacman has subfolders that could trip it up? But it's required for paru since they're in subdirs (~/.cache/paru/clone/packagename/)
- Updated Regex to find packages like `spotify 1:1.2.79.427-2`
- Auto-enabled git search since the search will show both cached builds and remote git builds (Cached on bottom)

src/downgrade
```diff
@@ -82,6 +83,28 @@ previously_installed() {
...
+  home=$(get_real_user_home)
+
+  [[ -z "$home" ]] && return 0
+
+  # yay: ~/.cache/yay/<pkgname>/
+  [[ -d "$home/.cache/yay" ]] && printf '%s\n' "$home/.cache/yay"
+  # paru: ~/.cache/paru/clone/<pkgname>/
+  [[ -d "$home/.cache/paru/clone" ]] && printf '%s\n' "$home/.cache/paru/clone"
+
+  return 0
```
pbrisbin: I guess this is no great complexity, but I didn't think we even needed that,


```suggestion  
  printf '%s\n' \
    "$home/.cache/yay" \
    "$home/.cache/paru/clone"
```

WDYT?

src/downgrade
```diff
@@ -207,11 +230,11 @@ search_git() {
     mapfile -t PACMAN_CACHE < <(read_pacman_conf CacheDir)
```
pbrisbin: I would just do it like this,

```suggestion
    mapfile -t PACMAN_CACHE < <(read_pacman_conf CacheDir; detect_aur_caches)
```

This is simpler, less error prone, and the way you have it currently means that passing `--cache` won't prevent the AUR caches from being appended, and I would expect it to.

pbrisbin (changes requested):

src/downgrade
```diff
@@ -204,14 +224,14 @@ search_git() {
...
+    mapfile -t PACMAN_CACHE < <(read_pacman_conf CacheDir; detect_aur_caches)
   fi
 
   for cache in "${PACMAN_CACHE[@]}"; do
-    path="$cache/$name" # directory name
+    path="${cache%/}/$name" # directory name
     if ! [[ -d "$path/.git" ]]; then
       {
-        gettext "\`$path\" not a git directory, skipping..."
+        gettext "\"$path\" not a git directory, skipping..."
```
pbrisbin: I think the intent here was probably the unixy tick-quote:


```suggestion
        gettext "\`$path' not a git directory, skipping..."
```

I doesn't it have to use some kind of eval to work properly with variables and localization? More reason just to ditch this warning I guess :)

src/downgrade
```diff
@@ -354,6 +378,16 @@ process_term() {
 
   candidates=($(printf '%s\n' "${candidates[@]}" | sort_packages))
 
+  # Auto-enable git search if an AUR cache has a git repo for this package
+  if ((!DOWNGRADE_FROM_GIT)); then
+    for cache in "${PACMAN_CACHE[@]}"; do
+      if [[ -d "$cache/$name/.git" ]]; then
+        DOWNGRADE_FROM_GIT=1
```
pbrisbin: Can we just / what would happen if, we remove this `DOWNGRADE_FROM_GIT` option and just always do that behavior if `$cache/$name/.git` exists?

i.e. just invert the conditional where the warning is at L232.

src/downgrade
```diff
@@ -355,9 +370,7 @@ process_term() {
   candidates=($(printf '%s\n' "${candidates[@]}" | sort_packages))
 
   # this is already sorted
-  if ((DOWNGRADE_FROM_GIT)); then
-    candidates=($(search_git "$name" | filter_packages "$name" "$operator" "$version") "${candidates[@]}")
-  fi
+  candidates=($(search_git "$name" | filter_packages "$name" "$operator" "$version") "${candidates[@]}")
```
pbrisbin: I wonder if this should've been up at L364. Something like,

```sh
candidates+=($({ search_cache "$name"; seach_git "$name"; } | filter_packages "$name" "$operator" "$version"))
```

Or even move `search_git` into `search_cache`. It seems like that behavior should only happen if `DOWNGRADE_FROM_CACHE` is set, as it is a _kind_ of cache search, no?

Since you're removing `--git` and making this "all the time", I think we may want to fix that, even if you didn't technically introduce it.

That said, you've done a lot here, so I'm OK to merge and fix after.

pbrisbin (approved):

test/aur_fallback/search_cache_aur.t
```diff
@@ -0,0 +1,43 @@
...
+search_cache finds packages in AUR helper cache directories
+
+  $ aur_cache=$(mktemp -d)
+  > mkdir -p "$aur_cache/foo"
+  > touch "$aur_cache/foo/foo-1.0-1-x86_64.pkg.tar.zst"
+  > DOWNGRADE_FROM_CACHE=1
+  > DOWNGRADE_MAXDEPTH=2
+  > DOWNGRADE_ARCH=x86_64
+  > PACMAN_CACHE=()
+  > _AUR_CACHE_DIRS=("$aur_cache")
```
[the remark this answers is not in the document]
ZorudaRinku: Of course I forgot to run tests again 💀

ZorudaRinku:
Hm so already-built cache sorting now gets thrown in (seemingly) randomly instead of just being appended on the bottom like it was before... We could do some sorting to sort them back to the bottom but I would think it would be more logical to just sort them by the version so it'd show the git version above/below the cache built version if they were the same version.
image

Sorting them perfectly by version number seems to require some trickery with parsing the non-epoch versions from git vs cache though. And I'm not certain all the possible version numberings that can happen...:

```
~/.cache/paru/clone/spotify/spotify-1:1.2.77.358-1-x86_64.pkg.tar.zst - Already built in cache
~/.cache/paru/clone/spotify/spotify-1.2.77.358-7f389b0-any.gitpkg.tar.gz - From git
``` 

Thoughts?

pbrisbin (approved):
I see about `pacsort`. I think using that and accepting whatever it does is the right move. If we think it's mishandling packages with epoch, that feels like a bug we could report there. I'm inclined to not care though.

pbrisbin:
I can take over to address the conflicts and Restyled failure. I think I also want to squash merge into one `feat:` commit. That OK with you?

ZorudaRinku:
Fine by me :) And yeah sorry I couldn't figure out the restyled thing lol

After thinking about this a lot more I think how I was imagining it was flipped. I don't think this is pacsorts fault, I think the problem is how we generated the list R276. Pacsort handles epochs fine, but when we were previously generating the git list we didn't include the epoch. And because I randomly decided to use spotify in testing it brought up how it was sorting poorly because of that. 

I think we should include the proper sorting I committed in [Epoch sorting](https://github.com/archlinux-downgrade/downgrade/pull/264/commits/37d722ab6acbd2e1bbce40a1006d37b5bcd0e7d4) as it should be the proper fix for this. It's just that the epoch is specified in a different variable in the pkgbuild instead of being included in the pkgver and it was just never accounted for.

PR #270, opened 2026-05-21
fix: check for XDG_CACHE_HOME

ShaiBlizzard:
### Checklist

* Admin:
  - [X] No duplicate PRs
* Integration:
  - [x] Always add unit tests for fixes or new functionality

### Description

Currently, if the user is using $XDG_CACHE_HOME which differs from $HOME/.cache, downgrade doesn't find anything
This makes it check both by default
(I've never worked with unit tests before so rather than mess up, set pr to draft)

### DIFF
--- a/test/aur_fallback/detect_aur_caches.t
+++ b/test/aur_fallback/detect_aur_caches.t
@@ -1,6 +1,5 @@
   $ source "$TESTDIR/../helper.sh"
   > PACMAN_LOG=/dev/null
-  $ unset XDG_CACHE_HOME
 
 Detects yay cache when present
 
--- a/test/aur_fallback/detect_aur_caches.t
+++ b/test/aur_fallback/detect_aur_caches.t
@@ -41,9 +41,6 @@ Detects both when both present, with xdg taking precedence
   > detect_aur_caches
   /tmp/*@xdg/yay (glob)
   /tmp/*@xdg/paru/clone (glob)
-Unset so doesnt get used by other tests
-  $ unset XDG_CACHE_HOME
-
 
 Returns nothing when neither present
 
--- a/test/aur_fallback/detect_aur_caches.t
+++ b/test/aur_fallback/detect_aur_caches.t
@@ -1,5 +1,6 @@
   $ source "$TESTDIR/../helper.sh"
   > PACMAN_LOG=/dev/null
+  $ unset XDG_CACHE_HOME
 
 Detects yay cache when present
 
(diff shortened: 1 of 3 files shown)


pbrisbin (approved):
Nice fix, thanks.

This feels tricky (and low value) to unit test so I wouldn't worry about it. As long as it doesn't break current tests.

test/aur_fallback/detect_aur_caches.t
```diff
@@ -1,5 +1,6 @@
   $ source "$TESTDIR/../helper.sh"
   > PACMAN_LOG=/dev/null
+  $ unset XDG_CACHE_HOME
```
pbrisbin: I think it be reasonable to clean the environment for all tests. So, you could add `export XDG_CACHE_HOME=` [here](https://github.com/archlinux-downgrade/downgrade/blob/91395e6f3d745c0c33f757c6bb3bf6df0780bf8d/test/helper.sh#L17-L23), for example.

Then your test below wouldn't need the extra cleanup.

pbrisbin (approved):

test/aur_fallback/detect_aur_caches.t
pbrisbin: > Not sure if unsetting it at first should go here or in helper.sh

Ah, sorry I just saw this. So yeah, I think I prefer that.

test/aur_fallback/detect_aur_caches.t
```diff
@@ -28,6 +28,23 @@ Detects both when both present
...
+  > mkdir -p "$fake_xdg/paru/clone"
+  > mkdir -p "$fake_home/.cache/yay"
+  > mkdir -p "$fake_home/.cache/paru/clone"
+  > get_real_user_home() { echo "$fake_home"; }
+  > detect_aur_caches
+  /tmp/*@xdg/yay (glob)
+  /tmp/*@xdg/paru/clone (glob)
+Unset so doesnt get used by other tests
+  $ unset XDG_CACHE_HOME
+
```
pbrisbin:
```suggestion
```

pbrisbin (approved):

test/aur_fallback/detect_aur_caches.t
```diff
@@ -28,6 +28,20 @@ Detects both when both present
...
+  $ fake_home=$(mktemp -d --suffix @home)
+  $ fake_xdg=$(mktemp -d --suffix @xdg)
+  > XDG_CACHE_HOME="$fake_xdg"
+  > mkdir -p "$fake_xdg/yay"
+  > mkdir -p "$fake_xdg/paru/clone"
+  > mkdir -p "$fake_home/.cache/yay"
+  > mkdir -p "$fake_home/.cache/paru/clone"
+  > get_real_user_home() { echo "$fake_home"; }
+  > detect_aur_caches
```
pbrisbin, line 41: My bad.
```suggestion
  > detect_aur_caches
  > XDG_CACHE_HOME=
```

pbrisbin (approved):

